PR #48, opened 2023-04-12
Fix geos website link

blackrez:
A little fix for the link to the new geos website. Is it plan to use an external website for the documentation ?


Maxxen:
Thanks! Yes, we should look into setting up something better for documenting all the functions etc. Or maybe we can do a GitHub wiki? We'll think about it.

PR #90, opened 2023-06-06
Add initial native OSM reader

Maxxen:
This PR adds an initial attempt at a native osm pdf reader/table function/replacement scan.

It will parse and decompress each file block in parallel and uses [protozero](https://github.com/mapbox/protozero) to scan the protobuf messages in a streaming fashion with minimal intermediate materialization. That said, there are still some inefficiencies we could improve upon.

Example:
```sql
DESCRIBE FROM './tmp/sweden.osm.pbf';
┌─────────────┬───────────────────────┬─────────┬─────────┬─────────┬─────────┐
│ column_name │      column_type      │  null   │   key   │ default │  extra  │
│   varchar   │        varchar        │ varchar │ varchar │ varchar │ varchar │
├─────────────┼───────────────────────┼─────────┼─────────┼─────────┼─────────┤
│ kind        │ OsmEntity             │ YES     │         │         │         │
│ id          │ BIGINT                │ YES     │         │         │         │
│ tags        │ MAP(VARCHAR, VARCHAR) │ YES     │         │         │         │
│ refs        │ BIGINT[]              │ YES     │         │         │         │
│ lat         │ DOUBLE                │ YES     │         │         │         │
│ lon         │ DOUBLE                │ YES     │         │         │         │
└─────────────┴───────────────────────┴─────────┴─────────┴─────────┴─────────┘

SELECT DISTINCT ON(kind) * FROM './tmp/sweden.osm.pbf' ;
┌───────────┬──────────┬──────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬────────────┬────────────────────┐
│   kind    │    id    │         tags         │                                                                                             refs                                                                                              │    lat     │        lon         │
│ osmentity │  int64   │ map(varchar, varch…  │                                                                                            int64[]                                                                                            │   double   │       double       │
├───────────┼──────────┼──────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼────────────┼────────────────────┤
│ Node │   566027 │                      │                                                                                                                                                                                               │ 57.8395566 │ 14.821418600000001 │
│ Way       │ 15511145 │ {highway=service, …  │ [5781735229, 126432394, 2542232911, 2542232917, 2542232927, 2542232929, 2542232933, 8953960346, 8953960347, 2542232935, 2542232941, 2542232944, 2542232967, 2542232976, 2542232980, 2542232…  │            │                    │
│ Relation  │  6080093 │ {addr:city=Borläng…  │ [60087961, 405667867]                                                                                                                                                                         │            │                    │
└───────────┴──────────┴──────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴────────────┴────────────────────┘
```

The primary limitation is that this table function will just scan and decode the raw osm pbf format, it will not actually reconstruct ways and relations into linestrings and polygons, but simply output them with their node references as duckdb lists.

For simple nodes it works fine though:
```sql
-- Give me the top 10 tallest named mountains in Sweden!
SELECT tags['name'][1], tags['ele'][1]::DOUBLE as ele 
FROM './tmp/sweden.osm.pbf' 
WHERE tags['natural'] = ['peak'] AND tags['name'] != [] AND kind = 'Node' 
ORDER BY ele DESC LIMIT 10;
┌───────────────────────┬────────┐
│    tags['name'][1]    │  ele   │
│        varchar        │ double │
├───────────────────────┼────────┤
│ Kebnekaise nordtoppen │ 2096.8 │
│ Kebnekaise sydtoppen  │ 2096.0 │
│ Sarektjåhkkå          │ 2089.0 │
│ Kaskasatjåkka         │ 2076.0 │
│ Nordtoppen            │ 2060.0 │
│ Sydtoppen             │ 2023.0 │
│ Áhkká                 │ 2011.0 │
│ Buchttoppen           │ 2010.0 │
│ Bårddetjåhkkå         │ 2005.0 │
│ Šielmmáčohkka         │ 2004.0 │
├───────────────────────┴────────┤
│ 10 rows              2 columns │
└────────────────────────────────┘
```

It is possible to manually perform a join with some sql-fu to e.g. construct linestrings out of ways, for example:
```bash
# launch with a disk-backed DB
duckdb tmp.db
```
```sql
CREATE TABLE ways AS FROM './tmp/sweden.osm.pbf' WHERE kind = 'Way';
CREATE TABLE nodes as FROM './tmp/sweden.osm.pbf' WHERE kind = 'Node';
-- Really gnarly SQL
SELECT way_id, ST_AsText(st_makeline(list(st_point(lat, lon)))) 
FROM (
    SELECT w.id as way_id, n.id as node_id, n.lat as lat, n.lon as lon 
    FROM (
        SELECT UNNEST(list_distinct(refs)) as ref, unnest(range(length(refs))) as idx, * 
        FROM (
            FROM WAYS where tags['highway'] != [])
        ) AS w 
    JOIN 
        nodes as n ON n.id = w.ref 
    ORDER BY w.id, w.idx ASC
) GROUP BY way_id;
```
But this is not as efficient as it could be and requires the entire result set to fit into memory. DuckDB has an amazing hash-aggregate operator but it doesn't support "out of core" execution entirely yet. (Although I think thats about to change soon).

Regardless, I'm thinking of how to make duckdb support this use case better and have this function go all the way and actually produce geometry types out of the box, but that's going to require a lot more effort, so I figured that getting this up at least may still be marginally useful in the meantime.


ttomasz:
@Maxxen  Sorry to write in closed PR but I don't quite get how reading relations would work with this solutions.

Relations can reference other entities like nodes, ways, and other relations.

As I understand fields refs is used in ways to get ids of nodes for a given way and for relations to get member objects. The problem is that to properly parse relations we need to know what kind of object is a member and what is its role.

See this relation: https://www.openstreetmap.org/relation/2730708#map=12/50.6789/17.9061

It has a bunch of members like:
- node 128054630 which points to the admin_centre
- bunch of ways that construct the geometry (outer roles)
- relation 3172807 which connects it to a subarea

With current implementation we end up with all ids as numbers in refs array like this:
image
which unfortunately makes it impossible to parse any relations that is more complicated than building with multipolygon or donut geometry.

Maxxen:
Hey @ttomasz, I've added both relation reference types and roles to the output in #108

PR #109, opened 2023-07-27
Apply patch and update duck

samansmink:
Applies outstanding patch fromDuckDB repo, switches to extension test loading instead of using the --test-dir. What this fixes is that the unittest binary that is produced in the build can now actually be run without the --test-dir param to run both standard DuckDB tests and the spatial tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,14 @@ endif
 CLIENT_FLAGS :=
 
 # These flags will make DuckDB build the extension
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks! I'd very much like if https://github.com/duckdb/duckdb/pull/8364 got merged before we bump the spatial version on the main repo though.

samansmink:
@Maxxen sure! after this PR is merged, we can bump spatial version in duckdb whenever suits you. The patch can/needs to be removed in that process

PR #114, opened 2023-08-11
Allow passing `HEADERS=[AUTO/FORCE/DISABLE]` st_read() to control xlsx behavior.

Maxxen:
This PR makes it possible to pass a `HEADERS=[AUTO/FORCE/DISABLE]` option to `st_read()` to control whether or not to read the first row as a header row or not when importing xlsx. We do this by setting a thread local config option programatically before opening the dataset, and resetting it after binding. Previously you would have to set this GDAL config option through an environment variable or an external configuration file, but now its possible to control it per-query and from within duckdb.

Example:
```sql
FROM ST_read('./output.xlsx', open_options = ['HEADERS=FORCE']);
```

Note that this workaround won't be necessary anymore since GDAL 3.8 (scheduled for November...?) as the XLSX options will be [possible to pass as open options](https://gdal.org/drivers/vector/xlsx.html) instead of configuration options natively.

### DIFF
--- a/spatial/src/spatial/gdal/functions/st_read.cpp
+++ b/spatial/src/spatial/gdal/functions/st_read.cpp
@@ -110,14 +110,26 @@ struct GdalScanLocalState : ArrowScanLocalState {
 
 struct GdalScanGlobalState : ArrowScanGlobalState {};
 
-struct ScopeGuard {
-	std::function<void()> f;
-	ScopeGuard(std::function<void()> f) : f(f) {
+
+struct ScopedOption {
+	string option;
+	string original_value;
+
+	ScopedOption(string option_p, const char* default_value) : option(option_p) {
+		// Save current value
+		original_value = CPLGetConfigOption(option.c_str(), default_value);
 	}
-	~ScopeGuard() {
-		f();
+
+	void Set(const char* new_value) {
+		CPLSetThreadLocalConfigOption(option.c_str(), new_value);
+	}
+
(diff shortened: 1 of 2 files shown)


davidjbrennan:
That's great news @Maxxen.  

Are you also able to include the only other configuration option "FIELD_TYPES=[STRING/AUTO]".  

This would make for a good work-around until the release of GDAL 3.8.

Maxxen:
You're right, good catch. I've added `FIELD_TYPES=[AUTO/STRING]` option as well now.

wonb168:
> This PR makes it possible to pass a `HEADERS=[AUTO/FORCE/DISABLE]` option to `st_read()` to control whether or not to read the first row as a header row or not when importing xlsx. We do this by setting a thread local config option programatically before opening the dataset, and resetting it after binding. Previously you would have to set this GDAL config option through an environment variable or an external configuration file, but now its possible to control it per-query and from within duckdb.
> 
> Example:
> ```sql
> FROM ST_read('./output.xlsx', open_options = ['HEADERS=FORCE']);
> ```
> 
> Note that this workaround won't be necessary anymore since GDAL 3.8 (scheduled for November...?) as the XLSX options will be [possible to pass as open options](https://gdal.org/drivers/vector/xlsx.html) instead of configuration options natively.

even use options，it still wrong for sheet1 in my excel in attachment，please try how to work for sheet1,tkx
!

PR #121, opened 2023-08-26
Apply patch from https://github.com/duckdb/duckdb/pull/8430 and update duckdb to latest main

carlopi:
I have only checked it compiles, nothing more involved, so a second look would be appreciated


Maxxen:
Tested it out locally and seems like nothings breaking. Thanks!

PR #128, opened 2023-09-12
Cache bounding boxes, add `st_extent`, `st_xmin/xmax/ymin/ymax` for all types, enable setting SRS of output geometry when writing through GDAL, optimize `st_intersects` joins

Maxxen:
### Cache bounding boxes
We now compute and store the bounding boxes of geometries upon serialization to the `GEOMETRY` type. This allows us to quickly get the extent and optimize a lot of operations by first testing if their bounding box overlaps before we spend time or memory on further deserialization and computation.

### Add `st_extent` and `st_xmin`, `st_xmax`, `st_ymin`, `st_xmin`
Because of the above these can now be implemented very easily and efficiently since they just have to deserialize the bounding box and nothing else, making it a o(n) operation for the `GEOMETRY` type. For the vectorized geometry `POLYGON_2D` and `LINESTRING_2D` types this is however computed on demand although there is no de/serialization costs for these anyway. For `BOX_2D` and `POINT_2D` these operations are basically zero-copy.

 ### Enable setting the output SRS when writing through GDAL
 You can now pass an optional "SRS" parameter to the GDAL copy function to set the spatial reference system of the output dataset explicitly. Example:

```sql
COPY (FROM t1) TO './test2.geojson' WITH (FORMAT GDAL, DRIVER 'GeoJSON', SRS 'EPSG:3857');
```
You should be able to pass [anything that OGR understands](https://gdal.org/api/ogrspatialref.html#_CPPv4N19OGRSpatialReference16SetFromUserInputEPKc)

### Optimize joins with `st_intersects` as join condition
Since getting the extents for geometries is so cheap now that we cache bounding boxes, I've implemented an experimental optimizer rule that rewrites inner joins using `st_intersects` as join condition to a comparison join on the extents followed by a st_intersects filter. This gets optimized into an `IE-join` which DuckDB does extremely well. I haven't tested it too much but I've seen a 5x improvement (joining 1mil nyc taxi ride pickup locations with 250~ taxi zones), and given that the complexity fundamentally changes (from O(MxN) to O(N log N)) I expect this to scale much better for even larger workloads.

I'm super excited about this and going to look into enabling this optimization for the other spatial predicates as well once I've done more testing.

[You can read more about DuckDB's IE joins here](https://duckdb.org/2022/05/27/iejoin.html#inequality-join-iejoin)


bretttully:
Cool! Will this enable predicate pushdown filtering on the bbox?

Maxxen:
Not quite yet, but that's next!

PR #129, opened 2023-09-17
[BROKEN] Steps towards being autoloadable

carlopi:
Mostly looking for feedback before going for the last bunch of changes.

Main question is if the ALTER_ON_CONFLICTS are actually needed or we can sort of safely assume that those names are ours to use.

This (plus another bunch of minor changes) would allow to move duckdb_spatial to be classified as autoloadable, not stricly needed but it might be cool to having it working without explicitly INSTALL and LOAD.


spatial/src/spatial/core/functions/scalar/st_ashexwkb.cpp
```diff
@@ -55,14 +56,12 @@ void GeometryAsHEXWKBFunction(DataChunk &args, ExpressionState &state, Vector &r
...
-
-	CreateScalarFunctionInfo info(ScalarFunction("ST_AsHEXWKB", {GeoTypes::GEOMETRY()}, LogicalType::VARCHAR,
+void CoreScalarFunctions::RegisterStAsHEXWKB(DatabaseInstance &instance) {
+	auto fun = ScalarFunction("ST_AsHEXWKB", {GeoTypes::GEOMETRY()}, LogicalType::VARCHAR,
 	                                             GeometryAsHEXWKBFunction, nullptr, nullptr, nullptr,
-	                                             GeometryFunctionLocalState::Init));
-	info.on_conflict = OnCreateConflict::ALTER_ON_CONFLICT;
-	catalog.CreateFunction(context, info);
+	                                             GeometryFunctionLocalState::Init);
+	// ?? info.on_conflict = OnCreateConflict::ALTER_ON_CONFLICT;
```
carlopi, line 63: @Maxxen: was this actually needed? or we can safely ignore this statement?

carlopi:
Understood why the ALTER_ON_CONFLICTS are there... now working around them.

carlopi:
@Maxxen: As a whole it changes semantic slightly, namely on conflicts where some function named ST_something exists it's not going to work properly. Internal conflicts have been resolved.

Unsure if that points to the fact that the mechanism in ExtensionUtil should allow for overload-sets (a general concern) and what's the way forward for spatial (= autoloadable or not?)

Maxxen:
@carlopi Hey, thanks for this! I'm not super happy that we have to cross the `geos` and `core` namespaces now, Ideally I would like to keep the different spatial "submodules" separate as much as possible. 

I'll merge this now though but we should add an overload to the ExtensionUtil so that it can take a `ON_CONFLICT` argument as well to allow different extensions to provide overloads to already existing functions. I think this is important for the extension ecosystem as a whole more than just spatial.

carlopi:
This PR https://github.com/duckdb/duckdb/pull/9065 looks like it's going to land soon, and that would allow to clean up the interdependencies between GEOS and CORE, allowing to implement this the proper way.

Later will update the PR accordingly.

Maxxen:
This has been implemented in #132

Issue #27, opened 2023-04-06
GeoParquet Support?


marklit:
This extension was compiled with GDAL 3.6.3 which has support for GeoParquet (it was added in 3.5.1). Any idea why it states for the format is unsupported?

```bash
$ /Volumes/Seagate/duckdb_spatial/build/debug/duckdb -unsigned test.duckdb
```
```sql
LOAD '/Volumes/Seagate/duckdb_spatial/build/debug/extension/spatial/spatial.duckdb_extension';
select * from st_read('/Volumes/Seagate/open5g_data/microsoft_roads_ai/Oceania_AUS.gpq') limit 1;
```

```
ERROR 4: `/Volumes/Seagate/open5g_data/microsoft_roads_ai/Oceania_AUS.gpq' not recognized as a supported file format.
Error: IO Error: Could not open file: /Volumes/Seagate/open5g_data/microsoft_roads_ai/Oceania_AUS.gpq (`/Volumes/Seagate/open5g_data/microsoft_roads_ai/Oceania_AUS.gpq' not recognized as a supported file format.)
```

The file itself looks fine.

```bash
$ ogrinfo microsoft_roads_ai/Oceania_AUS.gpq
````

```
INFO: Open of `microsoft_roads_ai/Oceania_AUS.gpq'
      using driver `Parquet' successful.
1: Oceania_AUS (Line String)
```

**Comment by Maxxen** [comment]
Yeah it is not supported yet. We already have our own Parquet reader extension in DuckDB and we are looking into how to integrate that with this extension in a natural way. I haven't tested it, but you should maybe be able to use the parquet extension to load the gpq and simply convert the wkb binary columns into geometries using `ST_GeomFromWKB` 

You can see the supported drivers using `SELECT * FROM st_list_drivers()`

**Comment by Maxxen** [comment]
I should explain: It is not supported because we don't bundle the Arrow library (which provides the parquet driver)

**Comment by marklit** [comment]
This might be unrelated to your extension but I got the following when trying to load the data in via the Parquet extension.

```
$ /Volumes/Seagate/duckdb_spatial/build/debug/duckdb -unsigned test.duckdb
```

```sql
install parquet;
load parquet;
```

```
[...]

PR #198, opened 2023-11-27
Feature/native gpq

aske-w:


aske-w:
Sorry, just ignore this PR.  Did not see I was not making the PR against my own repository :)

Maxxen:
Hi! Good work! This is an interesting approach, but it has the downside of coupling the spatial extension code to the parquet extension. 

Now that the latest DuckDB main contains the ability to [read/write parquet kv-metadata](https://github.com/duckdb/duckdb/pull/9126) my idea to eventually support GPQ would be to create a macro or plan replacement that attempts to bind or look up the `parquet_read` and `parquet_kv_metadata` in the catalog (and then add additional casts/operators) to avoid actually sharing any code with the parquet extension. If the parquet extension is not installed (and the initial catalog lookup/binding fails) we can just error.

aske-w:
Hi, thanks for the reply. I did not intend for this to end up in the spatial extension anyway 😅 This was for an assignment at universit. My benchmarking reveals this approach is slower and generally performs worse in terms of `perf stat` metrics than parquet + `st_geomfromwkb` mentioned in #27 :)

PR #206, opened 2023-12-08
Enable CI to compile for duckdb-wasm

carlopi:
Refactor of CI, unifying to extension-template, and add the necessary defines or patches to make so that it does cross-compile successfully.


Maxxen:
Super cool! We should look into maybe trying to upstream some of the patches in the future to make it less brittle, but very excited to merge this!

carlopi:
Thanks! Yes, PR is not super clean, but I wanted to get it over the line, so to be able to test also the system as a whole.

Very happy to make another round cleaning up cmake and patch duplication.

PR #227, opened 2024-01-10
Add `ST_Read_Meta`, don't throw in GDAL filesystem stat

### LINKED ISSUE #156: Provide a way to get the CRS data from `st_read`
### Discussed in https://github.com/duckdb/duckdb_spatial/discussions/155

<div type='discussions-op-text'>

<sup>Originally posted by **jdoig** October 25, 2023</sup>
Please excuse me if I'm being stupid (or just haven't read the manual thoroughly enough)... But is there anyway from within duckdb to get the CRS data from the `.prj` file in the shx I'm parsing with ST_READ?

I see that the extension integrates with PROJ which makes me suspect what I'm asking is possible (though I do see from the documentation that the geometry itself can't contain SRID information yet)</div>

-----
I see two approaches to handle this, we probably want both.
- Implement something like a `st_describe` to output the GDAL metadata from a dataset
- Output the CRS of the dataset as an additional columns when calling `st_read`, we just have to be careful to deduplicate the column names, or maybe even allow specifying the column name with a parameter.

Maxxen:
This PR adds a new `st_read_meta` table function to read the metadata present in files readable by st_read.
The output is similar to what you would get with `ogrinfo -json` but some fields are still missing. (like e.g. `metadata` ironically). Still, this finally allows you to inspect the projection info of a GDAL dataset layer from within DuckDB.

Closes #156

Also implements a small fix in the gdal virtual filesystem layer so that a failed stat call doesn't throw but instead returns -1.

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -283,8 +283,6 @@ select st_centroid('POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::geometry);
 -- POINT(0.5 0.5)
 ```
 
-End of example
-
 ## ST_Collect
 
 _Collects geometries into a collection geometry_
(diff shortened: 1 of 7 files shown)


cboettig:
@Maxxen is this available in the nightlies?  I'm trying to test with:

```sql
FORCE INSTALL 'spatial' FROM 'http://nightly-extensions.duckdb.org';
LOAD 'spatial'
```

But I still get the error: 

```
Error: Catalog Error: Table Function with name st_read_meta does not exist!
Did you mean "st_read"?
```

Maxxen:
What version are you on? Seems to work for me on v0.9.2
image

As well as on V0.10.0
image

cboettig:
my apologies! just updated and it's working now, (v0.9.2) thanks.

PR #249, opened 2024-02-04
Allow setting geometry type explicitly during layer creation, + misc Bugfixes

### LINKED ISSUE #233: COPY () TO WITH FORMAT OpenFileGDB failing
When using trying to write out to OpenFileGDB an error occurs.

```sql
COPY (select * from *.parquet') to 'admins.gdb' with (format GDAL, DRIVER 'OpenFileGDB');`
```

produces

```
ERROR 6: Unsupported geometry type
Error: IO Error: Could not create layer
```

More detail in this [discussion](https://github.com/duckdb/duckdb_spatial/discussions/229)

### LINKED ISSUE #238: Using st_read on File GeoDatabase results in error "No such file or directory"
Attempting to open a File Geodatabse using the `st_read()` function results in an error `Error: IO Error: GDAL Error (4): database.gdb not recognized as a supported file format.` This error occurs whether or not I specify the `layer` or `allowed_drivers` parameters. I am able to open a specific table by specifying a full path, e.g. `database.gdb\a0000000a.gdbtable` however I'd like to access the data using the path to the geodatabase and layername parameter.

I'm using the duckdb CLI version v0.9.1 on Windows.

### LINKED ISSUE #240: Column names with spaces fail in WHERE clause [.xlsx]
If you load a .xlsx where one of the column names has spaces, you can use it in a `SELECT` clause but not in a `WHERE` clause.

Repro: Use this excel file: 
[samsung_sales.xlsx](https://github.com/duckdb/duckdb_spatial/files/14069503/samsung_sales.xlsx)

Run:

`INSTALL spatial; LOAD spatial;`
`SELECT * FROM st_read('samsung_sales.xlsx', layer='Orders') WHERE \"Unit Selling Price\" > 14 LIMIT 10;`

or this script if preferred: 
[excel_col_space.py.zip](https://github.com/duckdb/duckdb_spatial/files/14069508/excel_col_space.py.zip)

Using DuckDB 0.9.2. Also did not work with installing nightly spatial.

Error: 
> Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result
Error: IO Error: GDAL Error (1): SQL Expression Parsing Error: syntax error, unexpected identifier, expecting end of string. Occurred around :
Unit Selling Price>14 AND Unit Selling Price 
     ^
[SQL: SELECT * FROM st_read('samsung_sales.xlsx', layer='Orders') WHERE "Unit Selling Price" > 14 LIMIT 10;]

### LINKED ISSUE #247: feature request: force casting when necessary
Hi,
I have these two files:

- https://gist.github.com/aborruso/d9ec192b5981d357c20e67d0ae341f7f
- https://gist.github.com/aborruso/0870135acca06966501814874b1bc2c4

If I do a JOIN

```
COPY (
    SELECT w.*, n.*
    FROM st_read('layer.geojson') w
    LEFT JOIN read_csv_auto('table.csv') n ON w.POSTAL = n."ISO_3166-1"
    ORDER BY POSTAL
) TO 'test_1.geojson' WITH (FORMAT 'GDAL', DRIVER 'GeoJSON')
```

I have this error

```
Error: near line 1: Not implemented Error: Unsupported field type
```

And I have it, because I have a `date` field in the CSV. 
If I remove the field `Data_Report`, I have no error.

I had to do some testing to figure out what the problem was, and this is an example with a few fields.
It would be optimal to have casting forcing for those fields not compatible with the output. An option that forces the field type to text, for non-compatible fields, without having to list them in hard coding.

I am not able to help with code writing, sorry. I report here simply my user experience.

Thank you

Maxxen:
- Adds support for writing TIME/DATE/TIMESTAMP etc columns through GDAL
- Some GDAL drivers will attempt to read from a file without first checking the file type, this broke when trying to read from directories as DuckDB would immediately throw. We now just return 0 bytes read instead, which triggers GDAL to actually inspect the file. This fixes being able to read from the directory path for formats that commonly span multiple files (such as .gdb and .shp), instead of having to find the explicit inner data files.
- You can now pass a geometry type parameter when copying through GDAL to create a geometry layer of a specific type, which is a requirement for some drivers (e.g. OpenFileGDB) as they don't support mixed geometries.
- Fixes a bug where column in filters that get pushed down into GDAL were not being properly quoted in the case that they contain spaces or special characters.

Example:

```sql
COPY (SELECT ST_Point(1,2) as geom, 10 as i) TO '__TEST_DIR__/test.gdb' WITH (FORMAT 'GDAL', DRIVER 'OpenFileGDB', GEOMETRY_TYPE 'POINT');
```

Closes #247 
Closes #240 
Closes #238 
Closes #233


aborruso:
great PR, thank you

davidjbrennan:
Hi @Maxxen.  I'm using a very recent build (v0.9.3-dev4116) of DuckDB and the issue is still present of a column name, containing spaces, not being properly quoted when the filter gets pushed down to GDAL using st_read against an Excel spreadsheet.  When is your fix likely to be included in the main build of DuckDB?

Maxxen:
Next DuckDB release is scheduled for tuesday!

PR #266, opened 2024-02-26
Fix wasm thread

carlopi:
This will have to be refactored, and that could happen while absorbing patches in deps/CMakeLists.txt

Idea of the changes is that in case of emscripten, threading support has to be added explicitly, and this needs to happen also at the C level.

sqlite_scanner has similar issue given that currently duckdb's CMakeLists.txt do not properly propagate C_FLAGS.

These changes are anyhow what it's needed to get spatial working with the `duckdb_thread` platform.


carlopi:
Note that currently (also to be fixed) duckdb-wasm CI for extensions is not performing any check, so CI will only check this compiles successfully.

PR #280, opened 2024-03-12
Support geometries with Z and M values

Maxxen:
This PR includes a major rework of the runtime geometry representation to enable support for Z and M values. Although no functions that specifically make use of the Z or M axis are implemented yet, and some existing functions may have limited support. Im not terribly happy with the vertex array abstraction yet but it works. Additionally:

There is a new GeometryProcessor abstraction which is used to unify deserialization to GEOS geometries and DuckDB native geometries while also enabling some operations to be performed _while_ deserializing, "zero-copy"-style so as to avoid materializing the geometry hierarchy entirely. The only functions to make use of this so far is `ST_Area`,  `ST_AsWKB`, `ST_(Min/Max)/X/Y/Z/M` and `ST_AsText`.

We now use our own WKT parser instead of relying on GEOS. I don't expect it to be more efficient but it does allow us to handle empty Z and M geometries and provide better error messages (eventually, they're not that exact yet).

New `ST_Force2D`, `ST_Force3DZ`, `ST_Force3DM` and `ST_Force4DM` functions have been added.


kylebarron:
Does this support Z and M in the native representations as well or just in the GEOS one?

Maxxen:
Native too! + WKB/GeoJSON/GDAL support. Although I haven't added new vectorized representations like `POLYGON_2D`

kylebarron:
> Although I haven't added new vectorized representations like POLYGON_2D

ah that was what I meant to ask!

PR #297, opened 2024-03-31
Add NearestPoints scalar function.

and-pulc:
Hi All,

This PR adds the [NearestPoints](https://libgeos.org/doxygen/geos__c_8h.html#a8bbe2ba198e5284b488b4c3ad1929113) function from the GEOS library as a scalar function DuckDB spatial. The GEOS C++ bindings don't seem to support a factory class for the MultiPoint type, so the resulting geometry is stored as a LineString. I've added the function to all the main documentation points I saw for other functions, but let me know if I missed anything!


Maxxen:
Hi! Thanks for the PR! Looks good, just some comments:

- We generally develop at the latest "stable" branch, currently `v0.10.1`, and only add stuff to main when we need to rely on changes in that only exists on main in the core DuckDB repository. Every week or so I'll backport/merge main with stable to keep them in sync.
- GEOS uses a common "collection" type for all multi/collection geometries, so you can create one using the `GEOSGeom_createCollection_r()` function, passing in `GEOS_MULTIPOINT` as the second argument. But linestring is also fine for now, I can fix it when I merge with main later.
- Regarding the docs, the "source of truth" is the separate markdown files in /docs, but this has completely changed on the stable branch anyway so this is fine for now too, ill move it over when I sync in a couple of days.

Thanks again for the PR! Feel free to reach out if you have any questions about the code.

Maxxen:
ah, the CI failures are unrelated, hold on and Ill fix it!

Maxxen:
Alright, I've switched to the stable branch instead, moved over the docs stuff and changed the name to `ST_ShortestLine` instead to be [more similar to what PostGIS does](https://postgis.net/docs/ST_ShortestLine.html). Ill work on merging main with stable separately so the changes here will be available in the nightly builds soon.

PR #318, opened 2024-05-16
Port changes from v0.10.2, apply patches from DuckDB core

Maxxen:


carlopi:
I think the solution to the _rtools failure and possibly also the one for wasm are here: https://github.com/carlopi/duckdb_spatial/pull/2, trying to see how that will end up failing.

PR #345, opened 2024-06-18
Enabling Android Builds

hannes:


hannes:
hang on `arm64-v8a` does not work yet

Maxxen:
The patch system for spatial is not very intelligent so it expects there to be a "zlib_wasm.patch" as well, which I think we can just make an empty file. I can push a commit to this PR tomorrow to fix it but otherwise looks good!

PR #375, opened 2024-08-21
Fix skipping of PrimitiveGroups in ST_ReadOSM

### LINKED ISSUE #349: Inconsistent row count between North America extract and planet file using `ST_ReadOsm()`
Hello!

I was playing around with `ST_ReadOsm()` and I noticed inconsistent results for the same query when run on the North America extract compared to the Planet file. I'm wondering if this might be a bug because I'm not sure how to explain the behavior.

The north america extract returns 245,000 rows but the planet only returns 90,000. I'd expect the planet to return far more than North America. Both files were downloaded today.

```
D SELECT COUNT(*) FROM st_readOSM('/home/wcedmisten/Downloads/planet-240610.osm.pbf') WHERE tags['amenity'] = ['restaurant'];
100% ▕████████████████████████████████████████████████████████████▏ 
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│        89751 │
└──────────────┘
D SELECT COUNT(*) FROM st_readOSM('/home/wcedmisten/Downloads/north-america-latest.osm.pbf') WHERE tags['amenity'] = ['restaurant'];
100% ▕████████████████████████████████████████████████████████████▏ 
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│       244678 │
└──────────────┘
```

DuckDB version:
```
~/Downloads/duckdb --version
v1.0.0 1f98600c2c
```

OS version:
```
lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
```

pilate:
Currently ST_ReadOSM is only reading the first 'PrimitiveGroup' per 'PrimitiveBlock'.

When the parsing loop seeks to the granularity/offset fields, it advances the block_reader position beyond any unprocessed groups, causing them to be missed.

By making a copy of block_reader, we can grab the granularity/offsets without messing up the real one.

This also makes sure we only have to read the values once per block.

Fixes #349.

### DIFF
--- a/spatial/src/spatial/core/io/osm/st_read_osm.cpp
+++ b/spatial/src/spatial/core/io/osm/st_read_osm.cpp
@@ -284,6 +284,20 @@ struct LocalState : LocalTableFunctionState {
 			string_table.push_back(string_table_reader.get_string());
 		}
 
+		// Need to read ahead without advancing block_reader
+		auto reader_copy = block_reader;
+
+		// Read the granularity and optional offsets
+		if (reader_copy.next(17)) {
+			granularity = reader_copy.get_int32();
+		}
+		if (reader_copy.next(19)) {
+			lat_offset = reader_copy.get_int64();
+		}
+		if (reader_copy.next(20)) {
+			lon_offset = reader_copy.get_int64();
+		}
+
 		state = ParseState::Block;
 	}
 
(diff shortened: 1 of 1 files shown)


Maxxen:
Hi! Thanks for digging in to this and fixing it, really appreciated! LGTM.

wcedmisten:
Thanks for the fix! Do you know when this fix would be released? Is there a preview version I can test with?

wcedmisten:
Never mind! available in the nightly build

https://github.com/duckdb/duckdb_spatial/issues/349#issuecomment-2335011589

PR #381, opened 2024-08-28
Upgrade dependencies

Maxxen:
GEOS 3.12 -> 3.13beta
GDAL 3.8 -> 3.9
PROJ 9.1.1 -> 9.4.1


Maxxen:
Note to self:
- Don't build shell on windows as the sqlite symbols conflict
- GDAL requires cpp17 headers that's not available on our CI

rcoup:
GDAL 3.10, GEOS 3.13, and Proj 9.5 are all out now... anything I can help with here?

FYI, there's also a change proposed for both [GDAL](https://gdal.org/en/latest/development/rfc/rfc102_embedded_resources.html) and [PROJ](https://osgeo-proj--4274.org.readthedocs.build/en/4274/community/rfc/rfc-8.html) to formalise embedding resources (like proj.db) into the libraries based on your existing approach.

Maxxen:
@rcoup Not really, we bumped to latest GEOS separately before DuckDB v1.1.1(I think..). We can't really move forward with latest GDAL because the c++17 requirement is not compatible with our CI setup, and ideally we want spatial to be available on older targets.

rcoup:
By "our CI environment" do you mean https://github.com/duckdb/extension-ci-tools ?

The `manylinux2014` toolchain has a newer GCC (v10) from Redhat such that it supports C++17, and there's newer GCC (v9/10/11/13) available from the Ubuntu team [for 18.04](https://launchpad.net/%7Eubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=bionic). So should be possible...

Maxxen:
> The manylinux2014 toolchain has a newer GCC (v10) from Redhat such that it supports C++17, and there's newer GCC (v9/10/11/13) available from the Ubuntu team [for 18.04](https://launchpad.net/%7Eubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=bionic). So should be possible...

CC @carlopi

carlopi:
Discussed with @Maxxen and @samansmink, I think this is something that would be interesting to look at this in general in the next release cycle.

Extensions (eg. spatial) are free to get ahead and experiment with this, but this comes with some cost, and maybe it's better to do this properly on the tooling side.

Maxxen:
We're moving to C++17 in the next major DuckDB release, so Ill see if I can update GDAL on the main branch soon.

PR #395, opened 2024-09-11
Fixup worflows for v1.1.0

carlopi:
Also remove unneded `.github/workflows/_extension_distribution.yml`, relying on extension-ci-tools.


carlopi:
Just for fun, I would expect this PR to fail compilation against main, given main now has the patch applied :)

But I would need this in first, then v1.1.0 tracking main (after this is in), then fixing patch.

PR #401, opened 2024-09-19
Fix example docs after moving sample files to different folder

ador:
Following the example docs led to IOerrors otherwise.


Maxxen (approved):

Maxxen:
Thanks!

PR #407, opened 2024-09-20
Serialize r-tree index scan and plan

Maxxen:
CC @Mytherin


Mytherin (approved):
Thanks for the PR, LGTM!

PR #429, opened 2024-10-14
Pull up pushed down filters from table scan before replacing with index scan

### LINKED ISSUE #428: INTERNAL Error: Attempted to access index 2 within vector of size 2 in WHERE when using spatial index
Hello,

I am getting the following error when my WHERE has a spatial query and another condition, when I don't return the value used in the other condition.  I'm using `v1.1.1 af39bd0dcf` of the CLI here, on a Mac running Sonoma 14.7:

```
SELECT id FROM segment WHERE subtype='road' AND ST_Intersects(geometry, ST_Buffer(ST_GeomFromText('POINT (-8476562 4795814)'), 100));

INTERNAL Error: Attempted to access index 2 within vector of size 2
```

If I do a `SELECT id, subtype FROM ...` it works as expected. 

To reproduce:
```
create table segment as select * from 's3://overturemaps-us-west-2/release/2024-09-18.0/theme=transportation/type=segment/part-00014-64fa5d4b-991b-493d-ad9c-41edcfbd4776-c000.zstd.parquet';
UPDATE segment SET geometry = ST_TRANSFORM(geometry, 'EPSG:4326', 'EPSG:3857', always_xy := true);
create index segment_geom_idx on segment USING RTREE (geometry);
```

Then, this will work:
```
SELECT id, subtype FROM segment WHERE subtype='road' AND ST_Intersects(geometry, ST_Buffer(ST_GeomFromText('POINT (-8476562 4795814)'), 100));
┌──────────────────────────────────┬─────────┐
│                id                │ subtype │
│             varchar              │ varchar │
├──────────────────────────────────┼─────────┤
│ 0862aac667ffffff047ffedcbc2db0f8 │ road    │
│ 0862aac667ffffff047de7f2111f86ad │ road    │
│ 0862aac667ffffff043df7e4c6756d14 │ road    │
└──────────────────────────────────┴─────────┘
```

but this will fail:
```
D SELECT id FROM segment WHERE subtype='road' AND ST_Intersects(geometry, ST_Buffer(ST_GeomFromText('POINT (-8476562 4795814)'), 100));
INTERNAL Error: Attempted to access index 2 within vector of size 2
This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/dev/internal_errors
```

Without creating the spatial index, it works, so some interaction there.

Maxxen:
Closes #428 

>~This is a bit of a band-aide fix that disabled index scans if the table already has normal filters pushed down. I think its generally expected that the index scan would be way more selective, but as we don't have the chance to swap in the index scan until after normal filters are already pushed down we can't really apply it instead.~

>~I made a quick attempt at trying to reconstruct and "pull up" the filters from the table scan again but it seems like it would be very fragile and error prone to do correctly 100% of the time, especially for more advanced queries. It's also a bit silly as we'd basically undo a lot of the work that went into pushing down the table filters to begin with and have to replace basically the whole logical get node.~

>~The proper solution would be to improve the index interface in DuckDB core to give custom indexes a chance to properly participate in the query planning process while filters are being evaluated for pushdown. We could then also look at stats/estimated cardinalities to decide whether or not to use an index scan or not, instead of just always matching specific predicates.~

We now pull up the filters from the table scan before replacing it with an index scan, ensuring index scans are applied even if other filters are present as well. This is usually the right thing to do as geometry predicates are usually much more expensive to evaluate compared to other filters that DuckDB can push down.


youngpm:
Awesome, thanks for the quick fix!  Guessing this will show up in v1.1.3?

Maxxen:
We're most likely going to hot-swap and update the extension we distribute for 1.1.2 tomorrow or early next week, so a `FORCE INSTALL spatial` should be all that's needed to get this fix once that's done. I just have another in-progress PR id like to sneak in first.

youngpm:
@Maxxen did this go out yet?  Just tried the force install but no luck!

Maxxen:
@youngpm Sorry we didn't bump the extension for v1.1.2, but this fix should be available in the recently released v1.1.3!

youngpm:
Ah thank you, got it!

PR #447, opened 2024-11-01
Very small example added for st_geometrytype

DrewScatterday:
When working with ST_GeometryType I was confused on what types the enum returned as well as proper usage of it

I won't be offended if you feel this is unnecessary as its a very small nit-picky change but just thought it would be helpful to add an example. Thanks for your hardwork!

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -1059,7 +1059,15 @@ ANY ST_GeometryType (col0 WKB_BLOB)
 
 #### Description
 
-Returns a 'GEOMETRY_TYPE' enum identifying the input geometry type.
+Returns a 'GEOMETRY_TYPE' enum identifying the input geometry type. Possible enum return types are: `POINT`, `LINESTRING`, `POLYGON`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON`, and `GEOMETRYCOLLECTION`
+
+#### Example
+
+```sql
+SELECT DISTINCT ST_GeometryType(ST_GeomFromText('POINT(1 1)'));
+----
+POINT  
+```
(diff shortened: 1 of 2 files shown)


Maxxen:
Hi! Thanks a lot for this PR and the help on improving the docs.
This should probably also be documented, but the source-of-truth for the function reference is in the c++ files themselves, the markdown is automatically generated by executing the `generate_function_reference.py` script in the root of the directory.

To add an example for e.g. ST_GeometryType you'd have to add it to the `DOC_EXAMPLE` string in `src/spatial/core/functions/scalar/st_geometrytype.cpp:90`. Then recompile and execute the generate_funciton_reference.py script.

If this seems like a lot of effort feel free to leave this PR open as is and I will try to pick it up and incorporate the changes next time im working on spatial!

We're planning to rework this in the future so that all functions are declared separately with their names/parameters/types/overloads/descriptions/examples in e.g. json files from which we can generate both the markdown docs and scaffold the c++ stubs from.

DrewScatterday:
No prob I'll look into this when I get chance 👍

DrewScatterday:
I updated the function doc but I need to troubleshoot my windows subsystem for linux setup so I can generate the build folder

DrewScatterday:
Okay whew, so I was able to setup my linux env with cmake, gcc, openssl, and ninja. I then was able to successfully do `make debug` (which took quite a while the first time which I think is expected since I was building everything) and then run `generate_function_reference.py`

The one odd thing is there a few function doc changes that I didn't add so I'm wondering if my fork wasn't up to date when I built or something? Let me know if I'm missing something @Maxxen

PR #448, opened 2024-11-02
improve function argument names in docs for a bunch of functions

zeen:
These now either match the source code (e.g., `ST_MakeEnvelope()`), or match the description in the docs (`geom1`, `geom2`).

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -496,8 +496,8 @@ select st_collectionextract('MULTIPOINT(1 2,3 4)'::geometry, 1);
 #### Signatures
 
 ```sql
-BOOLEAN ST_Contains (col0 POLYGON_2D, col1 POINT_2D)
-BOOLEAN ST_Contains (col0 GEOMETRY, col1 GEOMETRY)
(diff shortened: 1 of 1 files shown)


Maxxen:
Hi! Thanks a lot for this PR and the help on improving the docs. 
This should probably also be documented (ha), but the source-of-truth for the function reference is in the c++ files themselves, the markdown itself is generated by executing the `generate_function_reference.py` script in the root of the directory.

To add function parameter names for e.g. `ST_MakeEnvelope` you'd have to add 
```c++
DocUtil::AddFunctionParameterNames(db, "ST_MakeEnvelope", {"min_x", "min_y", "max_x", "max_y"});
```
 after the `RegisterFunction` call in e.g. `src/spatial/core/functions/scalar/st_makeenvelope.cpp:58
 
 We're planning to rework this in the future so that all functions are declared separately with their names/parameters/types/overloads/descriptions/examples in e.g. json files from which we can generate both the markdown docs and scaffold the c++ stubs from.

PR #477, opened 2025-01-13
Move all dependencies to VCPKG + apply patches and update DuckDB

Maxxen:


Maxxen:
Seems like everything works up until GDAL, which 
- won't build on mingw64
- won't build on wasm32
- transitively pulls in meson which requires a later python version not available on our manylinux container, 
- and probably won't build on gcc4 combat either due to c++17, but we didn't get that far this time.

PR #478, opened 2025-01-14
Retrigger CI on v1.1.3 to track down cross compilation failures

Maxxen:


Maxxen:
@carlopi Seems like arm cross compilation still works on v1.1.3. Im going to re-run it again but this time keep duckdb at 1.1.3 but use latest extension ci tools

Maxxen:
Cool, seems like that somewhere between CMake 3.21 and CMake 3.25 recursive-make cross compilation broke.

Maxxen:
PR is up https://github.com/duckdb/extension-ci-tools/pull/129

carlopi:
PR there has been merged, but then I remembered why the CMake bump was required in the first place: to have cmake work correctly for boost::filesystem, needed by the iceberg extension.

I am not sure what's worse.

We might have to have another proper look at this mess.

PR #485, opened 2025-01-18
Import/export geoarrow.wkb extension type from/to GEOMETRY

### LINKED ISSUE #153: Representation of spatial types on export to `ArrowArrayStream`
Currently when we export a spatial column to Arrow, we get:

```python
import duckdb
duckdb.sql("LOAD spatial;")
duckdb.sql("SELECT ST_GeomFromText('POINT (0 1)') as geom").to_arrow_table()
#> pyarrow.Table
#> geom: binary
#> ----
#> geom: [[000020000000000000000000010000000000000000000000000000000000F03F]]
```

With [geoarrow-python](https://github.com/geoarrow/geoarrow-python) we can register extension types to propagate CRS and type name through pyarrow machinery:

```python
import geoarrow.pyarrow as ga
array = ga.with_crs(ga.as_wkb(["POINT (0 1)"]), '{<some projjson>}')
array
#> GeometryExtensionArray:WkbType(geoarrow.wkb <{<some projjson>}>)[1]
#> <POINT (0 1)>
array.type.extension_name
#> 'geoarrow.wkb'
array.type.__arrow_ext_serialize__()
#> b'{"crs":"{<some projjson>}"}'
```

Would it be appropriate to export geometry columns to GeoArrow extension arrays (i.e., with `ARROW:extension:name` and `ARROW:extension:type` set according to https://github.com/geoarrow/geoarrow/blob/main/extension-types.md ? It looks like what is exported by default is your internal binary representation (perhaps not for the "native" unserialized types), but I think there are other types that are reencoded on export to Arrow (e.g., boolean columns are bitpacked).

paleolimbot:
This PR implements conversion between DuckDB GEOMETRY and geoarrow.wkb when importing or exporting Arrow using https://github.com/duckdb/duckdb/pull/15285 .

```
# build/duckdb -unsigned
v1.1.4-dev4808 79da908032
Enter ".help" for usage hints.
D LOAD nanoarrow;
D CALL register_geoarrow_extensions();
┌────────────┐
│ registered │
│  boolean   │
├────────────┤
│ true       │
└────────────┘
D COPY (SELECT st_geomfromtext('LINESTRING (0 1, 2 3)')) TO 'test.arrows' WITH (FORMAT ARROWS);
D SELECT * FROM "test.arrows";
┌──────────────────────────────────────────┐
│ st_geomfromtext('LINESTRING (0 1, 2 3)') │
│                 geometry                 │
├──────────────────────────────────────────┤
│ LINESTRING (0 1, 2 3)                    │
└──────────────────────────────────────────┘
```

Closes #153.


spatial/src/spatial/core/geoarrow.cpp
```diff
@@ -0,0 +1,117 @@
...
+		RegisterArrowExtensions(config);
+		output.SetValue(0, 0, true);
+	}
+
+	output.SetCardinality(1);
+	data.finished = true;
+}
+
+void GeoArrow::Register(DatabaseInstance &db) {
+	TableFunction register_func("register_geoarrow_extensions", {}, GeoArrowRegisterScan, GeoArrowRegisterBind);
```
Maxxen, line 111: Maybe this would be more idiomatic as a `PRAGMA` instead? Alternatively we could just always register on extension load.

spatial/src/spatial/core/geoarrow.cpp
```diff
@@ -0,0 +1,117 @@
+
+#include "spatial/core/geoarrow.hpp"
+
+#include "duckdb/common/arrow/arrow_converter.hpp"
+#include "duckdb/common/arrow/schema_metadata.hpp"
+#include "duckdb/function/table/arrow/arrow_duck_schema.hpp"
+#include "geos/vend/json.hpp"
```
Maxxen, line 7: We generally prefer yyjson, but I guess this is fine too

spatial/src/spatial/core/geoarrow.cpp
```diff
@@ -0,0 +1,117 @@
...
+		const auto options = context.GetClientProperties();
+		if (options.arrow_offset_size == ArrowOffsetSize::LARGE) {
+			schema.format = "Z";
+		} else {
+			schema.format = "z";
+		}
+	}
+
+	static void ArrowToDuck(ClientContext &context, Vector &source, Vector &result, idx_t count) {
+		CastParameters parameters;
```
Maxxen, line 59: Im not sure this is going to work? The cast parameters need to be properly instantiated so that we can access the arena allocator from inside WKBToGeometryCast. But doing that from "outside" the execution engine is a bit awkward. 

The cleanest (but least efficient) way would be to instantiate an ExpressionExecutor give it a BoundCastExpression, and let DuckDB figure out the rest. Alternatively we could just create a new arena and inline the vectorized WKB conversion there:

Something like
```c++

ArenaAllocator arena(BufferAllocator::Get(context));
// or just the default allocator (maybe more performant in this case, but no memory tracking)
// ArenaAllocator arena(Allocator::DefaultAllocator());

WKBReader reader(arena);

UnaryExecutor::ExecuteWithNulls<string_t, geometry_t>(
	  source, result, count, [&](string_t input, ValidityMask &mask, idx_t idx) {
		  try {
			  auto geom = reader.Deserialize(input);
			  return Geometry::Serialize(geom, result);
		  } catch (SerializationException &e) {
			 // Or just throw...?
			  mask.SetInvalid(idx);
			  return geometry_t {};
		  }
	  });
```

Ideally we wouldn't need an arena at all if we had a zero-alloc wkb->serialized geometry conversion like we do the other way around.

Maxxen:
Thanks a lot for picking this up! Some comments above, otherwise looks great!

spatial/src/spatial/core/geoarrow.cpp
paleolimbot: Either way! I couldn't figure out how to register a PRAGMA or a setting. If you're comfortable always registering this I don't envision any issues (I put the opt-in thing in case it was too close to release for this to make it in as a default setting).

spatial/src/spatial/core/geoarrow.cpp
paleolimbot: I'll update!

Maxxen:
Hey! Sorry, I saw that there were some activity that seemed related to this on the main repo so I kinda dropped my focus on it. I think everything looks good, but judging from the CI you might need to rebase and/or update the duckdb submodule to get CI to pass. But I can try to handle it tomorrow!

Maxxen:
@paleolimbot re: `st_geomfromwkb`, users are already getting confused by this when it comes to geoparquet/st_read conversion from WKB blobs to geometry... but my plan is to modify `st_geomfromwkb` to try to detect if the blob is actually a `geometry` type upon deserialization failure and explains the situation. Or just provide overload for the `GEOMETRY` type specifically that always throws a more descriptive error/explanation.

Maxxen:
@paleolimbot CI is green now so I think this is good to go. Im looking forward to playing around with it later. Thanks again for championing this!

PR #493, opened 2025-02-05
Bump to v1.2.0

smoy:
I followed pattern from https://github.com/duckdb/duckdb-delta/pull/148


smoy:
it's already working v1.2.0. no need to have this change.

PR #557, opened 2025-05-01
Duplicate symbols when statically linking avro and spatial

NiclasHaderer:
At MotherDuck, we build and statically link DuckDB extensions on our server. Since Iceberg has a dependency on Avro, we also have to like the Avro extension on the server. Unfortunately, this leads to a conflict for duplicate symbols, namely `json_object_get` when linking both _duckdb-spatial_ and _duckdb-avro_.

This issue stems from the inclusion of the `jansson` library in avro and the `gdal` library in spatial.

Both libraries define a `jsob_object_get` function (see https://github.com/akheron/jansson/blob/96d160df90016066d04d493d1d69639474ba4f20/src/jansson.h#L189 and https://github.com/OSGeo/gdal/blob/9c6652c257799b88ca83302c3368545aba869c65/ogr/ogrsf_frmts/geojson/libjson/json_object.h#L348).

---

This PR patches the `json_object_get` function in `gdal` by renaming it to `gdal__json_object_get`.

### DIFF
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -107,7 +107,7 @@ ExternalProject_Add(
     DEPENDS ${GDAL_DEPENDENCIES}
     URL ${CMAKE_CURRENT_SOURCE_DIR}/vendor/gdal385.zip
     CONFIGURE_HANDLED_BY_BUILD TRUE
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks for the PR! Dont worry about failing CI, I need to apply patches from DuckDB main, but I can take it from here!

NiclasHaderer:
Thanks a lot

NiclasHaderer:
Hey. Is it possible to get this PR shipped before the RC? It is currently preventing us from building Iceberg and Spatial together

Maxxen:
Yes, I got it merged on my dev branch and am gonna bump main soon. I got distracted by trying to look into solving the root cause here and have been investigating stripping symbols, etc, but that will have to wait until another time.

NiclasHaderer:
Sounds good, thanks

PR #568, opened 2025-05-16
Fixes connected to duckdb-wasm AND bump vcpkg-duckdb-ports

carlopi:
Some Wasm related changes, long overdue.
1. link in more libraries
2. move expat to vcpkg-duckdb-ports
3. CI to target v1.3-ossivalis

This should have no effects on regular targets, but Wasm targets should increase significantly to about 23MB.
I will take a look (once CI completes) at the resulting binaries / eventual failures.


Maxxen:
Thanks! MinGW failure is spurious, so ill just merge this now

Issue #591, opened 2025-05-30
`ST_Affine()` doesn't handle 3D transformation correctly


yutannihilation:
The below example moves the point `(1, 1)` by `(2, 3, 0)`. So, the result should be `(3, 4)`. Actually, PostGIS returns `(3, 4)` with the same query.

```sql
D SELECT ST_AsText(ST_Affine(ST_Point(1, 1)::geometry,
ﾂｷ                            1, 0, 0,   -- a, b, c
ﾂｷ                            0, 1, 0,   -- d, e, f
ﾂｷ                            0, 0, 1,   -- g, h, i
ﾂｷ                            2, 3, 0)); -- xoff, yoff, zoff
┌────────────────────────────────────────────────────────────────────────────────────────────┐
│ st_astext(st_affine(CAST(st_point(1, 1) AS geometry), 1, 0, 0, 0, 1, 0, 0, 0, 1, 2, 3, 0)) │
│                                          varchar                                           │
├────────────────────────────────────────────────────────────────────────────────────────────┤
│ POINT (4 1)                                                                                │
└────────────────────────────────────────────────────────────────────────────────────────────┘
```

Also, `ST_Affine()` cannot handle `POINT_3D`.

```sql
D SELECT ST_AsText(ST_Affine(ST_Point3D(1, 1, 0),
ﾂｷ                            1, 0, 0,   -- a, b, c
ﾂｷ                            0, 1, 0,   -- d, e, f
ﾂｷ                            0, 0, 1,   -- g, h, i
ﾂｷ                            2, 3, 0)); -- xoff, yoff, zoff
Binder Error:
No function matches the given name and argument types 'ST_Affine(POINT_3D, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL, INTEGER_LITERAL)'. You might need to add explicit type casts.
        Candidate functions:
        ST_Affine(GEOMETRY, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE, DOUBLE) -> GEOMETRY
[...]

PR #592, opened 2025-05-30
Add documents for `ST_Rotate*()`, `ST_Translate()` etc

### LINKED ISSUE #590: Some functions are missing on the function reference
It seems functions added by https://github.com/duckdb/duckdb-spatial/pull/532 are not documented in `docs/functions.md`.

ST_Polygonize
ST_Node
ST_MaximumInscribedCircle
ST_Affine,
ST_RotateZ,
ST_RotateX,
ST_RotateY,
ST_Scale,
ST_Translate,
ST_TransScale

yutannihilation:
Close #590 

Note that this pull request is based on the diff generated by Claude Code, but I manually verified all the results (and found some strange behaviors I reported in https://github.com/duckdb/duckdb-spatial/issues/591). So, if this contains some mistake, it's my own fault.

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -5,6 +5,7 @@
 
 | Function | Summary |
 | --- | --- |
+| [`ST_Affine`](#st_affine) | Apply an affine transformation to a geometry |
 | [`ST_Area`](#st_area) | Compute the area of a geometry. |
 | [`ST_Area_Spheroid`](#st_area_spheroid) | Returns the area of a geometry in meters, using an ellipsoidal model of the earth |
 | [`ST_AsGeoJSON`](#st_asgeojson) | Returns the geometry as a GeoJSON fragment |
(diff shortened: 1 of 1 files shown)


Maxxen:
Thanks for the PR! Improving documentation is always very appreciated!
`spatial` has a bit of a weird workflow with docs though, the "source of truth" for the docs is embedded in the c++ files, and registered for the functions by using the `FunctionBuilder`. This makes the docs available from inside DuckDB, I.e. you can see the function description when running: `SELECT * FROM duckdb_functions`. We then generate the markdown by running the `generate_function_reference.py` script, which basically queries the functions from DuckDB. 

The idea behind this is that it's "easier" to keep the docs updated if it's closer to the code, and since the markdown is generated by what's inside of DuckDB the content of the two won't diverge. However we might change this whole system in the future - Im not super happy with how it works in practice, and it makes it a lot harder for outside contributors that just want to help out with the docs.

yutannihilation:
It seems macros cannot have tags. So, probably the SQL in the `generate_function_reference.py` script needs some other trick.

```
D SELECT function_name, tags FROM duckdb_functions() as funcs where function_type = 'macro' and tags['ext'] is not null;
┌───────────────┬───────────────────────┐
│ function_name │         tags          │
│    varchar    │ map(varchar, varchar) │
├───────────────┴───────────────────────┤
│                0 rows                 │
└───────────────────────────────────────┘
```

cf. https://github.com/duckdb/duckdb/blob/4dc61eb5e38809554afa39b4b7c97547c0dee8c6/src/include/duckdb/function/macro_function.hpp#L36-L41

yutannihilation:
I gave up addressing these macros variants of `ST_Affine()` in this pull request. This pull request just adds a document to `ST_Affine()` and re-run the generate_function_reference.py script.

yutannihilation:
I changed my mind and add some tweak to `generate_function_reference.py` to generate "Macro function" section. Ideally, macro functions should belong to either the individual section for scalar function etc, but I couldn't find a nice way. Probably, listing them is better than nothing.

One remaining issue is that, if a macro has multiple definitions, the description and the example is `NULL` (e.g. `ST_Scale`, `ST_Translate`). Probably, this needs some fix on DuckDB API's side.

docs/functions.md
```diff
@@ -2748,6 +2865,143 @@ Computes the union of a set of input geometries
...
 
+## Macro Functions
+
+### ST_Rotate
+
+
+#### Signature
+
+```sql
+None ST_Rotate (geom None, radians None)
```
yutannihilation, line 2876: Hmm...

Maxxen:
Ahh sorry about that, you should be able to add  `duckdb_extension_load(json)` to the `extension_config.cmake` to tell DuckDB to build json, or also add `CORE_EXTENSIONS=json,parquet` to the `Makefile` to also include it in the build.

As for the macros, thats also an oversight, we should probably patch this in the future so that we can tag macros properly... but I think the work-around is good!

Maxxen:
You can do it just to generate the docs locally, but ideally you don't commit the inclusion of json so we don't spend time building it in CI unnecessarily.

PR #595, opened 2025-05-31
Add 3D types (`LINESTRING_3D`, `POLYGON_3D`)  and more conversions

### LINKED ISSUE #582: Feature request: Add `LINESTRING_3D` and `POLYGON_3D` types
Usually, users don't need to be aware of the spatial types (e.g. whether the column is `POINT_2D` or `GEOMETRY`). However, I found I need to care it when I export the query result as GeoArrow data. For example, if it's `GEOMETRY`, the exported column is `BinaryArray` when it needs to be `List<Struct<x: double, y: double>>` (LineString). It might be that such data conversion is supposed to happen in some GeoArrow libraries, not in DuckDB, but I'm not really sure.

e.g. https://github.com/yutannihilation/maplibre-geoarrow-deckgl-layers/blob/main/generate_lines.sql

In the case of 2D, I could just cast `GEOMETRY` to `LINESTRING_2D` or `POLYGON_2D`. However, it seems there's no luck if it's a 3D type. So, probably it would be good if spatial extension provides a type alias and the conversions to the type for 3D cases. I think the code would be something like this (I'm yet to figure out how to build duckdb-spatial, so I'm not sure if this works...):

https://github.com/yutannihilation/duckdb-spatial/compare/main...yutannihilation:add-3d-types

yutannihilation:
Close #582

- Add `LINESTRING_3D`
- Add `POLYGON_3D`
- Add conversions between 3D types and GEOMETRY
- Add conversions from 3D types to VARCHAR (to display)
- Add conversions from 3D types to 2D types (no opposite conversion)


yutannihilation:
This pull request is ready for review now. At first, I was planning to add conversions between the 2D type and the 3D type. But, considering this pull request is already a bit huge, I decided not to include it here. I hope I can address in another pull request.

Maxxen:
Thanks! Looks good!
Good job hacking around the whole casting code, there's still a bunch of remnants from before we did a bunch of refactoring (I.e. `CoreVectorOperations`, throwing instead of failing casts properly) that we should do another pass over in the future.

I think having implicit casts from `_3D` to `_2D` (and the other way around) would be great - one of the main ideas behind storing the coordinates in separate struct vector arrays is that casting between mixed dimensions is extremely cheap (you can just reference the vectors you care about, don't have to do any copying). But I agree that it can come in a future PR!

yutannihilation:
> having implicit casts from `_3D` to `_2D`

Sure, I will add it! What I wasn't sure was the 2D-to-3D conversion.

yutannihilation:
Thanks for reviewing!

PR #624, opened 2025-06-26
Update MainDistributionPipeline.yml: branches, no tags

carlopi:
I think problem is that on `v1.3.1` (and likely on other branches as well), extension is never actually deployed (given it's not a tag and it's not a branch called `main`.

This attempts at fixing this.

CI here basically only checks the syntax.


yutannihilation:
Thanks, I confirmed `FORCE INSTALL spatial FROM core_nightly;` works now!

PR #625, opened 2025-06-26
Update duckdb to v1.3.1

yutannihilation:
I think the problem is that currently this links v1.3.1 of DuckDB. If I try to load the debug built of current v1.3.1 branch, it fails. This pull request updates the commit to v1.3.1 tag of duckdb repo.

```
duckdb -unsigned -c "LOAD 'build/debug/extension/spatial/spatial.duckdb_extension';"
```

`Invalid Input Error:`
`Failed to load 'build/debug/extension/spatial/spatial.duckdb_extension', The file was built specifically for DuckDB version 'v1.3.0' and can only be loaded with that version of DuckDB. (this version of DuckDB is 'v1.3.1')`

edit:
The CI setting is correctly referring to v1.3.1, so this might not matter on CI builds. But, anyway I think updating this is a good idea.

https://github.com/duckdb/duckdb-spatial/blob/v1.3.1/.github/workflows/MainDistributionPipeline.yml


Maxxen:
Thanks!

PR #628, opened 2025-06-30
Formatting fixes for SQL examples

szarnyasg:


yutannihilation:
Thanks for catching the differences that I missed in https://github.com/duckdb/duckdb-web/pull/5544!

According to https://github.com/duckdb/duckdb-spatial/pull/593#issuecomment-2927015177, the base branch should be `v1.3.1` instead of `main`.

Also, you can generate `docs/functions.md` by running `python3 generate_function_reference.py`, but probably this step is optional (building duckdb-spatial might require some efforts).

PR #633, opened 2025-07-05
Disable '&&' op for now

Maxxen:
We can't currently add overloads for functions that span extensions when their signature contains custom types (in this case `BOX_2D`). This needs to be fixed in `duckdb/duckdb` but since we got a new release on Monday I'm disabling this for now so that we can merge the other bug fixes in the meantime.

### DIFF
--- a/docs/functions.md
+++ b/docs/functions.md
@@ -5,7 +5,6 @@
 
 | Function | Summary |
 | --- | --- |
-| [`&&`](#&&) | Returns true if the bounding boxes intersects. |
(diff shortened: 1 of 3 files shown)


yutannihilation:
Oh, sorry for this. I didn't notice the problem...

Maxxen:
No problem, I didn't realize this was going to be a problem either! I don't think it's going to be that hard to fix, I know who worked on this last, but it will have to wait until after v1.3.2.

PR #665, opened 2025-08-21
Add ST_ClipByRect

bradh:
This currently falls back to a GEOS implementation for anything more complex than points. I had originally intended to do the linestring implementation too, but it became ugly. 

This is a step in the implementation for the MVT geometry conversion, but wanted to avoid that getting too complicated, and it seems like useful functionality as a standalone function.

I'm not especially happy with the "GeosOperations" [bridge code](https://github.com/duckdb/duckdb-spatial/compare/main...bradh:duckdb-spatial:clip_rect_2025-08-08?expand=1#diff-6a0aec8e4545f6f4b93dfa6e42acafe141c2a84d29cd835d3c9264482daad3b6R2776-R2796), but unsure of a better way. Would definitely appreciate review comments on that.

### DIFF
--- a/src/sgl/sgl.cpp
+++ b/src/sgl/sgl.cpp
@@ -1,12 +1,15 @@
 #include "sgl.hpp"
 
 #include <algorithm>
+#include <cassert>
 #include <cstdlib>
 #include <cstring>
 #include <cmath>
+#include <iostream>
 #include <new>
 #include <queue>
 #include <string> // TODO: Remove this, it is only used for error messages in WKT reader
+#include <vector>
 
 //======================================================================================================================
 // Helpers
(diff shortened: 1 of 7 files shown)


yharby:
Thanks for taking this on. Quick question on current state, are you still planning to land it? It shows as conflicting now, happy to help test a rebase. On the GeosOperations bridge you flagged, the wrapper `get_clipped` already exists in `geos_geometry.hpp` calling `GEOSClipByRect_r`, so one option is a thin `ST_ClipByRect` struct that calls it directly, the same way `ST_Simplify` wraps `get_simplified`, which might let the bridge layer go away. Does that fit what you had in mind?

bradh:
Given its been so long with no attention, probably better just closed.

PR #683, opened 2025-09-19
ST_CoverageClean() - adding initial support

tbbuck:
Hi there! I hope you've been keeping well since I last dropped a PR on you out of the blue :) Amazing work on the recent Spatial updates lately btw, I've been constantly blown away by the table join join performance we now have at our fingertips thanks to you :-D

This PR implements 3 functions for the `CoverageClean` functionality recently added to `GEOS`:

* `GEOMETRY ST_CoverageClean (geoms GEOMETRY[], snapDistance DOUBLE, maxWidth DOUBLE)`
* `GEOMETRY ST_CoverageClean (geoms GEOMETRY[], snapDistance DOUBLE)`
* `GEOMETRY ST_CoverageClean (geoms GEOMETRY[])`

Somewhat bizarrely, I've been unable to replicate the [CoverageCleanerTest.cpp](https://github.com/libgeos/geos/blob/c05da7dacd42ad55f77bc7ef980e53bab5a012a8/tests/unit/coverage/CoverageCleanerTest.cpp#L20) tests from libgeos with either this of extension or the `geosop` tool. The extension code does align with output from `geosop` though.

Future nice-to-haves:

* Support `OverlapMergeStrategy`: I couldn't work out how to implement a 4-parameter function call in the same style as the other GEOS functions. I also got a bit concerned about unintended consequences of using enums from other packages.
* Improve my VCPKG version knowledge. I've hacked in the version change as an override to `vcpkg.json` - feels like a nasty bodge job.
* Implement aggregate `ST_CoverageClean_Agg()` versions of the above.

I'm very happy to work on any or all of the above with you btw, if you can share any guidance with me I'm all ears on making this function fly.

Thanks again!


Maxxen:
Hello! Thanks for the PR!

This is currently failing because the VCPKG baseline DuckDB's CI build system uses doesn't include the latest GEOS version.  I'll see if we can have a look at bumping it soon at which point this PR should be able to pass CI.

tbbuck:
Hello! Ahhhh that makes sense - I was struggling to figure out what was going wrong where, thanks for signposting it :) Is there anything I can do on that side that means less work for you?

For accepting an additional 4th parameter, I've just realised that I could call `SenaryExecutor` and ignore the extra columns (i.e. no need to find or implement a `QuateraryExecutor`). Do you think that's a good way for me to add an extra parameter here?

Thanks again!

maxmalynowsky:
Hi @tbbuck and @Maxxen, I've been a huge fan of your work until now. These DuckDB capabilities have allowed the teams I work with to move from PostGIS to DuckDB for many of our big data ETL pipelines, and DuckDB-WASM is making it possible for us to build very use-case specific web apps that broaden the accessibility of geospatial processing to more less-technical audiences across the world.

One of the tools I've been trying to port from PostGIS to DuckDB-WASM lately is a topology modification tool: https://em.fieldmaps.io/ (https://github.com/fieldmaps/edge-matcher). It does really complex data transformations, but it absolutely depends on clean input topology to work. I've been struggling to help make topology cleaning more accessible to people I know who only use ArcGIS and QGIS. Being able to use `ST_CoverageClean()` in DuckDB would be one of the most impactful things to happen for me in 2026, and I'd be happy to help get this along to the finish line in any way I can.

Maxxen:
We've bumped our vcpkg version so this should be able to make it in for next release. I'll try to have a look at getting this merged soon.

Maxxen:
Merged in https://github.com/duckdb/duckdb-spatial/pull/810

PR #692, opened 2025-10-23
fix invalid geometries in ST_AsMVTGeom

Jesperpaulsen:
Changes GEOS_PREC_NO_TOPO to GEOS_PREC_VALID_OUTPUT in get_gridded() to ensure topological validity during grid snapping operations.

The current ST_AsMVTGeom sometimes produces invalid geometries because the topology is not preserved.

Screenshot 2025-10-23 at 09 32 31
Screenshot 2025-10-23 at 09 32 28

### DIFF
--- a/src/spatial/modules/geos/geos_geometry.hpp
+++ b/src/spatial/modules/geos/geos_geometry.hpp
@@ -369,7 +369,7 @@ inline GeosGeometry GeosGeometry::get_transformed(const double matrix[6]) const 
 }
 
 inline GeosGeometry GeosGeometry::get_gridded(double grid_size) const {
-	return GeosGeometry(handle, GEOSGeom_setPrecision_r(handle, geom, grid_size, GEOS_PREC_NO_TOPO));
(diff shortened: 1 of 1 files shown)


Jesperpaulsen:
> Hey! Thanks for opening this PR. Don't worry about the CI failure, I'll take care of it.
> 
> I didn't use `GEOS_PREC_VALID_OUTPUT ` because AFAIK it might cause GEOS to collapse invalid geometries into lower dimensions (e.g. polygons to lines, and lines to points) and I wasnt sure how well visualization libraries would handle MVT tiles with mixed geometry types in practice. But I think this change might be preferable, lets see if we get any complaints in the future 🤷

Ah, yeah, that makes sense! Haven't worked to much with GEOS, but that's probably true. When it comes to the visualisation of mixed geometry types it shouldn't be any problem with Maplibre and Mapbox at least. It support multiple types in one source and its also possible to filter the layers based on what geometry types you want to show in the layer 😄

Issue #583, opened 2025-05-26
Feature Request: ST_Subdivide


sambcom:
Just like the nicely formatted https://github.com/duckdb/duckdb-spatial/issues/367 I'd love to add [ST_Subdivide](https://postgis.net/docs/ST_Subdivide.html) to avoid a copy back and forth to Postgres.

The new spatial join stuff is fantastic! I found with humongous polygons with 100s of thousands of points I can speed up ST_Intersects joins by over 50x by using [ST_Subdivide](https://postgis.net/docs/ST_Subdivide.html) to balloon the row count and decrease the complexity of each polygon.

JoelJaeschke:
Hey there, did not have as much time as I thought so far. Sorry if this is a stupid question, but how much memory is required to build DuckDB? I have 16GB on my laptop and whenever I run `make debug`, the process gets OOM-killed. Is this expected or am I doing something wrong? The normal build running just `make` works though. I have tried switching the ld for mold, but same thing happens.

jakub-ga:
hi @JoelJaeschke, thank you for implementing that exact feature. Do you plan to merge it soon?

PR #719, opened 2025-11-29
Add ST_Subdivide

JoelJaeschke:
This commit addresses #583 and adds ST_Subdivide, similar to how PostGIS does it. The algorithm does not reproduce PostGIS exactly, however. Every type of input geometry is accepted, except for GeometryCollections. The result is a GeometryCollection of subdivided geometry parts. The user can specify the maximum number of vertices that every part should have.

Slight additions to GeosGeometry were necessary, namely:
- Initialize new geometry from bounding rectangle
- Expose some more of the GEOS geometry's attributes
- Add a function that counts number of vertices of a geometry

### DIFF
--- a/src/spatial/modules/geos/geos_geometry.hpp
+++ b/src/spatial/modules/geos/geos_geometry.hpp
@@ -16,6 +16,7 @@ class GeosGeometry {
 public:
 	// constructor
 	GeosGeometry(GEOSContextHandle_t handle_p, GEOSGeometry *geom_p);
+	GeosGeometry(GEOSContextHandle_t handle_p, double xmin, double ymin, double xmax, double ymax);
 
 	// disable copy
 	GeosGeometry(const GeosGeometry &) = delete;
@@ -41,6 +42,7 @@ public:
 	bool is_valid() const;
 	bool is_empty() const;
 
+	GeosGeometry get_clone() const;
 	GeosGeometry get_boundary() const;
 	GeosGeometry get_centroid() const;
 	GeosGeometry get_convex_hull() const;
@@ -91,6 +93,7 @@ public:
 	// default tolerance is max(height/width) / 1000
 	GeosGeometry get_maximum_inscribed_circle() const;
 	GeosGeometry get_point_n(int n) const;
+	GeosGeometry get_geometry_n(int n) const;
 
 	GeosGeometry get_linemerged(bool directed) const;
 	GeosGeometry get_concave_hull(const double ratio, const bool allowHoles) const;
@@ -108,6 +111,13 @@ public:
 	void get_extent(double &xmin, double &ymin, double &xmax, double &ymax) const {
 		GEOSGeom_getExtent_r(handle, geom, &xmin, &ymin, &xmax, &ymax);
 	}
+	size_t get_dimension() const {
+		return GEOSGeom_getDimensions_r(handle, geom);
+	}
+	size_t get_num_geometries() const {
+		return GEOSGetNumGeometries_r(handle, geom);
+	}
+	size_t get_num_vertices() const;
 
 private:
 	GEOSContextHandle_t handle;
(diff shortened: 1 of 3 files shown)


JoelJaeschke:
Hey @Maxxen could you give me a pointer on why CI failed? This does not look like anything I changed, so I am a bit confused whether I caused that. Also, should I merge into main or the v1.4 branch? Thanks!

Maxxen:
Hey, sorry this got buried in my backlog. Im gonna try to review this soon in the coming days.

The main branch sometimes lags behind upstream duckdb main which is used in the CI workflow and so these sort of unrelated errors can surface. Generally v1.4-andium branch is what will be used to build spatial for the v1.4-series of patch releases, but main is used for the next upcoming minor release (v1.5 in this case), so I guess it depends if your PR needs to make use of upcoming features or if you want it in the next patch version. 

But now v1.4 is so close to release anyway I would just keep it on main. I think it should be enough to rebase otherwise I can have a look and try to fix-up this PR later so it can pass CI.

> The algorithm does not reproduce PostGIS exactly, however.

Exact PostGIS behavior is not strictly a requirement, but I somehow have a feeling I'm the one thats going to have to fix this once a user opens an issue about getting different results. Perhaps it's better if we can limit/explicitly throw an error in the harder-to-mimic cases if the semantics differ?

JoelJaeschke:
No worries, and thanks for taking the time!

I think it makes sense to keep it on main. I was just confused as main already used the native geometry type whereas v1.4 still used the geometry type from duckdb-spatial I believe.

Regarding the difference versus PostGIS, PostGIS chooses the node closest to center to do the split, whereas the version I implemented just splits the geometry along the geometric center, irrespective of whether there is a node there or not. The advantage of the PostGIS approach is that it does not (possibly) introduce two new nodes per split and thus may finish faster, but the algorithm becomes more complex. I fully understand your argument, so if you would prefer to have the PR use the same approach as PostGIS, I would happily implement that.

gsueur:
Could this be added to the DuckDB main version ? This would be very helpful and I guess the hard work has already been done. It's a pity to not see it shared with the community !

JoelJaeschke:
Hello @Maxxen this somehow also got burried in my inbox and I since totally forgot about it. I rebased on the current v1.5 branch and made two small adjustments. Your review would be greatly appreciated and happy to make changes as required.

MurrayData:
I would really appreciate the work that's going into adding `ST_Subdivide` as I'm working with large, complex shape flood risk zone polygons with huge numbers of points.

At present I'm running a workaround using the DuckDB Postgres extension and PostGIS running on ephemeral fast NVMe storage in a sidecar container:

```    CREATE OR REPLACE TABLE rofrs_sub AS
    SELECT * exclude geometry_wkb,
    ST_SetCRS(ST_GeomFromWKB(geometry_wkb), 'EPSG:27700') AS geometry
    FROM postgres_query(
        'pg',
        '
            SELECT 
                "Risk_band", "Confidence", "Shape_Length", "Shape_Area",
                ST_AsBinary(sub_geom.geometry) AS geometry_wkb      -- Appends the new, subdivided chunks
            FROM 
                subdivide_source AS s
            CROSS JOIN LATERAL 
                ST_SubDivide(s."Shape", 4096) AS sub_geom(geometry);
        '
    );
```
    
Which works fine and brings back 24 million sub divided geometries to DuckDB, from 382,000 original, but adds a significant overhead. I am saving the subdivided table as geoparquet for future use. 

A query of 42 million UK UPRN points in polygons in DuckDB then only takes 8 seconds compared to several hours without this.

Ultimately this will allow us to eliminate PostGIS from our workflows. We went serverless a few years ago and now use sidecars occasionally with ephemeral storage when DuckDB doesn't yet support what we're trying to do or performance is too slow at present.

Thanks.

Maxxen:
Okay, I'll try to review this soon, and if not figure something out before 2.0 at the very latest.

Maxxen:
Ah you know what lets just go for it, this will be in v1.5.6

gsueur:
Hi @Maxxen if you could as well take a look at https://github.com/duckdb/duckdb/pull/22076 which has been closed because stale ? I still think Geoparquet v1.1 is of interest for the Geospatial community and as I will be presenting at CNG2026 and FOSS4GNA, I would have liked to promote it then, instead of having the default v1.0 format.
Thanks

MurrayData:
> Ah you know what lets just go for it, this will be in v1.5.6

Thank you very much, as I have the PostGIS based workflows in situ already, I'll feed back comparisons.

Maxxen:
@gsueur will do

PR #725, opened 2025-12-09
Minor fixes for DuckDB-main

Maxxen:


Maxxen:
~~Hey @staticlibs do you recognize this windows arm CI error? This PR contains your patch from the v1.4-andium branch, but it seems like I still fail on GDAL.~~

Nvm, its failing on the x86 build so I must have borked something

Issue #718, opened 2025-11-27
Missing accessor to get rings of a rather complex (multi)polygon


pinkerltm:
I can access the ST_ExteriorRing() of a polygon, but there is currently no accessor to get interior rings like PostGIS [ST_InteriorRingN()](https://postgis.net/docs/ST_InteriorRingN.html) or [ST_DumpRings()](https://postgis.net/docs/ST_DumpRings.html).

This is needed for more complex intersections and topological analysis.

PR #726, opened 2025-12-09
Add ST_InteriorRingN

pinkerltm:
Please accept my very first contribution where I add ST_InterioRingN as described in #718

### DIFF
--- a/src/spatial/modules/main/spatial_functions_scalar.cpp
+++ b/src/spatial/modules/main/spatial_functions_scalar.cpp
@@ -3463,8 +3463,9 @@ struct ST_ExteriorRing {
 			    }
 
 			    if (geom.is_empty()) {
-				    const sgl::geometry empty(sgl::geometry_type::LINESTRING, geom.has_z(), geom.has_m());
-				    return lstate.Serialize(result, empty);
+				    // Polygon empty -> return empty linestring
+				    sgl::geometry empty_linestring(sgl::geometry_type::LINESTRING, geom.has_z(), geom.has_m());
+				    return lstate.Serialize(result, empty_linestring);
 			    }
 
 			    const auto shell = geom.get_first_part();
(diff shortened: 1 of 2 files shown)


src/spatial/modules/main/spatial_functions_scalar.cpp
```diff
@@ -6256,6 +6257,239 @@ struct ST_Hilbert {
...
+        BinaryExecutor::Execute<string_t, int64_t, string_t>(
+            args.data[0], args.data[1], result, args.size(),
+            [&](const string_t &blob, const int64_t &n) {
+                sgl::geometry geom;
+                lstate.Deserialize(blob, geom);
+
+                // ---- validate geometry ----
+                if (geom.get_type() != sgl::geometry_type::POLYGON) {
+                    // same semantics as ST_ExteriorRing: return NULL for non-polygons
+                    return string_t{};
```
Maxxen: This isn't how you return null, you need to use a separate executor `BinaryExecutor::ExecuteWithNulls` and set the null bit in the validity mask. See e.g. the implementation of `ST_Azimuth` for a reference on how to use it.

src/spatial/modules/main/spatial_functions_scalar.cpp
```diff
@@ -6256,6 +6257,239 @@ struct ST_Hilbert {
...
+                if (n < 1) {
+                    // invalid index → NULL
+                    return string_t{};
+                }
+
+                const idx_t num_parts = geom.get_part_count(); // includes shell
+                // parts: 0 = exterior, 1..n = interior rings
+                const idx_t num_interior = num_parts > 0 ? num_parts - 1 : 0;
+
+                if ((idx_t)n > num_interior) {
```
Maxxen: Minor nitpick, but we generally prefer c++-style `static_cast` or `UnsafeNumericCast<>` over c-style casts

src/spatial/modules/main/spatial_functions_scalar.cpp
```diff
@@ -6256,6 +6257,239 @@ struct ST_Hilbert {
...
+			if (n_format.validity.RowIsValid(n_format.sel->get_index(i))) {
+				nr = n_data[n_format.sel->get_index(i)];
+			} else {
+				n_is_null = true;
+			}
+
+			if (n_is_null) {
+				FlatVector::SetNull(line_vec, i, true);
+				continue;
+			}
```
Maxxen: Maybe this logic can be cleaned up a bit, something like:

```c++
const auto n_idx = n_format.sel->get_index(i);
if(!n_format.validity.RowIsValid(n_idx)) { 
  FlatVector::SetNull(line_vec, i, true);
  continue;
}

const auto nr = n_data[n_idx];
....
```

src/spatial/modules/main/spatial_functions_scalar.cpp
```diff
@@ -6256,6 +6257,239 @@ struct ST_Hilbert {
...
+
+			const idx_t ring_count = poly.length;
+			const idx_t interior_count = (ring_count > 0 ? ring_count - 1 : 0);
+
+			if (nr < 1 || nr > static_cast<int64_t>(interior_count)) {
+				// out of range -> NULL result
+				FlatVector::SetNull(line_vec, i, true);
+				continue;
+			}
```
Maxxen: Do we always return an empty linestring if the polygon is empty, even if the index is out of bounds? Would it be more consistent to check that the index is valid first? What does e.g. PostGIS do?

test/sql/geometry/st_interiorringn.test
```diff
@@ -0,0 +1,48 @@
...
+ ) AS t(geom);
+----
+LINESTRING (0 0, 1 0, 1 1, 0 1, 0 0)
+LINESTRING EMPTY
+LINESTRING (0 0, 2 0, 1 1, 0 2, 0 0)
+
+
+endloop
+
+
```
Maxxen: Can we add some tests for edge-cases, e.g. when n is out of bounds, or input contains nulls?

Maxxen (changes requested):
Hello! Thanks for the PR. Looks mostly good, but left some comments.

src/spatial/modules/main/spatial_functions_scalar.cpp
pinkerltm: ah! Thanks for the hint

src/spatial/modules/main/spatial_functions_scalar.cpp
pinkerltm: Well, you are right... for an POLYGON EMPTY the index is always out of bounds, even for n=0. So PostGIS always returns NULL for these cases which makes sense when I think about it. 

So I will handle these in the check for a valid index first and also return NULL then.

src/spatial/modules/main/spatial_functions_scalar.cpp
pinkerltm: fixed

test/sql/geometry/st_interiorringn.test
pinkerltm: The testfile was broken anyway. I fixed it and added tests for the negative cases as well. feel free to request more if needed.

PR #732, opened 2026-01-04
fix ring orientation for ST_AsMVTGeom and add orientation tests

### LINKED ISSUE #731: ST_AsMVTGeom outputs CCW exterior rings in tile coords
I ran into a problem with tiles generated via `ST_AsMVT`/`ST_AsMVTGeom` where they get rejected by strictly spec compliant MVT decoders (e.g., tippecanoe-decode, QGIS) with errors like `Polygon begins with an inner ring` or no error / no visible features in case of QGIS. I've to say that this isn't an issue with webapp renderers like MapLibre GL or similar clients, because they seem to be more relaxed.

After digging into the generated MVT tiles, I found that `ST_AsMVTGeom` isn't returning the right ring orientation when transforming from geo coordinates to tile coordinates.

## What is going wrong

According to the [Mapbox Vector Tile Spec](https://mapbox.github.io/vector-tile-spec/), ring orientation in tile (screen) coordinates where Y points down is defined as:

* Exterior rings must be clockwise (CW)
* Interior rings (holes) must be counter-clockwise (CCW)

Looking at the code, I think the problem is that `ST_AsMVTGeom` flips the Y axis (Y‑up → Y‑down), then calls `orient_polygons(true)`. However, `orient_polygons(true)` enforces CW in math coordinates (Y‑up), which becomes CCW after the Y‑flip in tile coordinates. As a result, exterior rings end up CCW in tile coordinates, which violates the MVT spec.

## Minimal Example

Here is a small repro:

```sql
D SELECT ST_AsText(
      ST_AsMVTGeom(
          ST_GeomFromText('POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))'),
          ST_Extent(ST_GeomFromText('POLYGON((0 0, 2 0, 2 2, 0 2, 0 0))')),
          4096
      )
  );
┌───────────────────────────────────────────────────────────────────────────────┐
│ st_astext(st_asmvtgeom(st_geomfromtext('POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))'…  │
│                                    varchar                                    │
├───────────────────────────────────────────────────────────────────────────────┤
│ POLYGON ((2048 4096, 2048 2048, 0 2048, 0 4096, 2048 4096))                   │
└───────────────────────────────────────────────────────────────────────────────┘
```

In Y-down tile coordinates, this polygon is CCW, even though it represents an exterior ring. Strictly spec-compliant MVT decoders therefore interpret it as an interior ring. 

I’ve tested a fix for this, and it resolves the issue with the strict decoders like QGIS and tippecanoe (the tiles load correctly there).

hafenkran:
Switched `ST_AsMVTGeom` to orient polygons the right way after the Y‑flip, so exterior rings are CW and holes are CCW in tile coords (as per the MVT spec). Added tests covering a polygon, a polygon with a hole, and a multipolygon to lock this in.

Example:
```sql
D SELECT ST_AsText(
      ST_AsMVTGeom(
          ST_GeomFromText('POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))'),
          ST_Extent(ST_GeomFromText('POLYGON((0 0, 2 0, 2 2, 0 2, 0 0))')),
          4096
      )
  );
┌──────────────────────────────────────────────────────────────────────────────────────────┐
│ st_astext(st_asmvtgeom(st_geomfromtext('POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))'), st_exten…  │
│                                         varchar                                          │
├──────────────────────────────────────────────────────────────────────────────────────────┤
│ POLYGON ((2048 4096, 0 4096, 0 2048, 2048 2048, 2048 4096))                              │
└──────────────────────────────────────────────────────────────────────────────────────────┘
```

Closes #731

### DIFF
--- a/src/spatial/modules/geos/geos_module.cpp
+++ b/src/spatial/modules/geos/geos_module.cpp
@@ -366,7 +366,7 @@ struct ST_AsMVTGeom {
 			if (!bind_data.clip) {
 
 				// But first orient in place
-				snapped.orient_polygons(true);
+				snapped.orient_polygons(false);
 
 				res_data[out_idx] = lstate.Serialize(result, snapped);
 				continue;
@@ -389,7 +389,7 @@ struct ST_AsMVTGeom {
 			auto cleaned_clipped = clipped.get_gridded(1.0);
 
 			// Also orient the polygons in place
-			cleaned_clipped.orient_polygons(true);
+			cleaned_clipped.orient_polygons(false);
 
 			res_data[out_idx] = lstate.Serialize(result, cleaned_clipped);
 		}
--- a/test/sql/mvt/st_asmvtgeom.test
+++ b/test/sql/mvt/st_asmvtgeom.test
@@ -1,0 +1,51 @@
+# name: test/sql/mvt/st_asmvtgeom.test
+# description: Test that ST_AsMVTGeom produces correct ring orientation according to MVT spec
+# group: [mvt]
+
+require spatial
+
+# Verify that ST_AsMVTGeom produces CW exterior rings in tile coordinates
+# 
+# Input: CCW Polygon in geographic coordinates (Y-up)
+# Expected: CW Polygon in tile coordinates (Y-down)
+# MVT Spec: "Exterior rings must be oriented clockwise [...] (when viewed in screen coordinates)"
+query I
+SELECT ST_AsText(
+    ST_AsMVTGeom(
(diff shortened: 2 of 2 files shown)


Maxxen:
Thanks for the PR! LGTM!

Maxxen:
(don't worry about the windows failure, I will have to deal with it myself soon anyway)

PR #737, opened 2026-01-10
Pin extension_ci_tools to 1.4 in 1.4 branch

staticlibs:
Due to the integration of duckdb/duckdb#19415 and
duckdb/extension-ci-tools#276 the `1.4-andium` branch of Spatial extension can no longer be built with `main` branch of `extension-ci-tools`.

This PR pins the version of `extension-ci-tools` used for CI builds. It should fix the following build problem on Windows:

```
error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease'
```

Ref: duckdblabs/duckdb-internal#2036

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -25,20 +25,20 @@ concurrency:
 jobs:
   duckdb-latest-build:
     name: Build extension binaries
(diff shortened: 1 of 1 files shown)


Maxxen:
Thanks a ton!

PR #743, opened 2026-01-18
Add debug build to spatial CI

carlopi:
Attempt at checking whether this separate configuration would help catch CI issues.

Open questions:
1. does it work, as in actually expand what's tested?
2. does it make so the uploaded extensions will be mixed up?

Second one is the most problematic, since it can introduce a new class of issues, to be reviewed properly after CI goes through.


Maxxen:
Do we have any idea why the CI fails? It seems like the musl run is not using VCPKG properly?

Maxxen:
We now run relassert builds instead, so closing this.

Issue #728, opened 2025-12-16
GDAL incorrectly says "file does not exist" when ran concurrently


C-Loftus:
I have a simple program which runs many concurrent read only operations with duckdb using the [go client](https://github.com/duckdb/duckdb-go)

However every once in a while I get this issue:

```
IO Error: GDAL Error (4): /Users/cloftus/github/project/my_file.fgb: No such file or directory
```

I have tried on both local and remote in gcs and both have the same issue (feel free to test against that file if desired; it is public open data)

```
IO Error: GDAL Error (4): Failed to open file gcs://national-hydrologic-geospatial-fabric-reference-hydrofabric/reference_catchments_and_flowlines.fgb,: {"exception_type":"HTTP","exception_message":"Unable to connect to URL \"https://storage.googleapis.com/national-hydrologic-geospatial-fabric-reference-hydrofabric/reference_catchments_and_flowlines.fgb%2C\": 404 (Not Found).","header_Access-Control-Allow-Origin":"*","header_Date":"Tue, 16 Dec 2025 14:27:44 GMT","header_Content-Type":"application/xml; charset=UTF-8","header_Alt-Svc":"h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000","header_Expires":"Tue, 16 Dec 2025 14:27:44 GMT","header_X-GUploader-UploadID":"AHVrFxMHB0OjZv81bDHRqWvbKAHCvOh1aXMUa49-4hyz0KxhY3cfLLrOp-NrgY3LaAqZiVREKl8xfz4","header_Cache-Control":"private, max-age=0","response_body":"","reason":"Not Found","header_Access-Control-Expose-Headers":"Content-Type","header_Server":"UploadServer","header_Content-Length":"261","status_code":"404"}

LINE 3:                 FROM ST_Read(
             ^ 
exit status 1
```

I am confused why I am getting this error since:
- All my operations are read only in the same process
- I have checks before the SQL that ensures the file exists
- I don't update the file or change the connection concurrently
- My path is absolute and I don't change my current working directory at all 
- It only happens around 30% of the time and has no clear pattern 

Do I need to add a mutex on this operation even though it is read only? 

---

# Code Context

If it is relevant, this is the entire go code which runs the SQL which causes the error. I have no other duckdb interactions in my codebase at the moment

Sorry if it is verbose; the actual SQL for duckdb is quite brief.

```go
// This function I only open once in my program
[...]

PR #773, opened 2026-03-17
Improve spatial filter pushdown in `ST_Read` + fix rtree scan issue

### LINKED ISSUE #769: RTREE index is not used
### What happens?

If I create a table with a geometry column and an id column RTREE index is not used:
When, if I create the table with only the geometry column, then the RTREE index is used.

### To Reproduce

```
INSTALL spatial;
LOAD spatial;

CREATE TABLE IF NOT EXISTS t1 (
            id          LONG PRIMARY KEY,
            geom        GEOMETRY
);


-- Create a table with 10_000_000 random points
INSERT INTO t1(id, geom)
SELECT row_number() OVER (), point
FROM st_generatepoints({min_x: 0, min_y: 0, max_x: 100, max_y: 100}::BOX_2D, 10_000, 1337);

-- Create an index on the table.

-- Perform a query with a "spatial predicate" on the indexed geometry column
-- Note how the second argument in this case, the ST_MakeEnvelope call is a "constant"
SELECT count(*) FROM t1 WHERE ST_Within(geom, ST_MakeEnvelope(45, 45, 65, 65));

-- RTREE is not used
EXPLAIN SELECT count(*) FROM t1 WHERE ST_Within(geom, ST_MakeEnvelope(45, 45, 65, 65));
```

### OS:

MacOS

### DuckDB Version:

1.5

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Maurizio Merli

### Affiliation:

Xthema

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

Maxxen:
This PR makes the spatial filter pushdown into `ST_Read` more robust so that it can handle cases where the filtered table column is wrapped in additional expressions (such as a cast).

We now also enable bounding-box pushdown for all spatial predicates, but only _remove_ the filter entirely if its `&&` or `ST_Intersects_Extent`.

Also implements `to_string` for `ST_Read` so that the layer index and any pushed down bbox filter is printed as part of the `EXPLAIN` output.

Also fixes a bug causing R-Tree index scans to not be used if there is another index type on the table (such as a primary key).

Closes #769 
Closes [#728](https://github.com/duckdb/duckdb-spatial/issues/728#issuecomment-4071376986)

### DIFF
--- a/src/spatial/index/rtree/rtree_index_plan_scan.cpp
+++ b/src/spatial/index/rtree/rtree_index_plan_scan.cpp
@@ -218,7 +218,7 @@ public:
 
 		unordered_set<string> spatial_predicates = {
 		    "ST_Equals",   "ST_Intersects", "ST_Touches",   "ST_Crosses",          "ST_Within", "ST_Contains",
-		    "ST_Overlaps", "ST_Covers",     "ST_CoveredBy", "ST_ContainsProperly", "&&",        "ST_IntersectsExtent"};
+		    "ST_Overlaps", "ST_Covers",     "ST_CoveredBy", "ST_ContainsProperly", "&&",        "ST_Intersects_Extent"};
 
 		table_info.BindIndexes(context, RTreeIndex::TYPE_NAME);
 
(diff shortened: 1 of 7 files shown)


dashie:
from which version it is available the fix?

Maxxen:
This should be available on v1.5.1 (released yesterday)

Issue #729, opened 2025-12-20
Please standardize file access and basic capabilities


quillaja:
Various file formats should be able to be opened using the same syntax. Also `ST_Read_Meta` fails in various ways on different formats.

System: 
- using cli: `DuckDB v1.4.3 (Andium) d1dc88f950`; Windows 11 amd64/x64
- using api: DuckDB.NET v1.4.3 on .NET 10

(geo)parquet - `ST_Read_Meta` **crashes**
```sql
FROM 'some_data.parquet'; -- works
FROM ST_Read('some_data.parquet'); -- does NOT work, error
FROM read_parquet('some_data.parquet'); -- works
```

geopackage - `ST_Read_Meta` works
```sql
FROM 'some_data.gpkg'; -- works
FROM ST_Read('some_data.gpkg'); -- works
```

geodatabase - `ST_Read_Meta` does not crash but does return 0 rows
```sql
FROM 'some_data.gdb'; -- does NOT work, error
FROM ST_Read('some_data.gdb'); -- works
```

shapefile - `ST_Read_Meta` works
```sql
FROM 'some_data.shp'; -- works
FROM ST_Read('some_data.shp'); -- works
```

Maxxen:
Hello!
This is expected behavior. 
The `parquet` reading capabilities are not part of `spatial`, but the `parquet extension, and so require different functions to access.

ST_Read itself wraps GDAL, but the underlying format drivers that GDAL use all vary widely in their capabilities, and so making sure they all output the same information from e.g. `ST_Read_Meta` is difficult.

As for the shorthand `select * from 'string.file_format'` syntax, again, we only support that for a handful of specific formats. GeoDatabase should maybe be added there, but the GeoDatabase driver is itself not super smooth IMO and so you probably want to pass some options when reading through it, and so are better off calling `ST_Read`.

https://duckdb.org/docs/stable/core_extensions/spatial/functions#st_read

> The following formats are currently recognized by their file extension:

|Format | Extension|
|-- | --|
|ESRI ShapeFile | .shp|
|GeoPackage | .gpkg|
|FlatGeoBuf | .fgb|

Issue #544, opened 2025-04-14
calling st_read_meta() crash duckdb


nmatton:
When calling the st_read_meta function, duckdb crash silently and no output is returned.

**How to reproduce :**

1/ Creating a sample parquet file:

```python
import geopandas as gpd
from shapely.geometry import Point, Polygon
print(f"Création d'un fichier GeoParquet d'exemple : {file_path}")
data = {
    'id': [1, 2, 3, 4],
    'nom': ['Paris', 'Lyon', 'Parc_Naturel', 'Zone_Inondable'],
    'type': ['Ville', 'Ville', 'Parc', 'Zone'],
    'population': [2141000, 513275, None, None],
    'geometry': [
        Point(2.3522, 48.8566), # near Paris
        Point(4.8357, 45.7640), # near Lyon
        Polygon([(5.0, 44.0), (6.0, 44.0), (6.0, 45.0), (5.0, 45.0)]), 
        Polygon([(1.0, 48.0), (3.0, 48.0), (2.5, 49.5), (1.5, 49.0)]) 
    ]
}
gdf = gpd.GeoDataFrame(data, crs="EPSG:4326") # WGS84
gdf.to_parquet(file_path)
```

2/ call the `st_read_meta` function on a duckdb instance :

```bash
 ~/>duckdb
v1.1.3 19864453f7
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D INSTALL spatial;
D LOAD spatial;
D SELECT * FROM st_read_meta('sample_data.parquet');
 ~/>
```

Maxxen:
ST_Read_Meta doest not work with parquet files. You should use the `parquet_metadata` or `parquet_kv_metadata` functions. 

Regardless, this should probably not crash. Ill look into it!

Issue #45, opened 2023-04-11
How to query a layer of OpenFileGDB dataset?


aborruso:
Hi,
I have a multilayer OpenFileGDB file. Its name is simple.gdb.
If I run

```
select * from st_read('simple.gdb') limit 10;
```

I have the first ten records. How to read what's the layer name of these 10 records?

How to query only the layer "lorermIpsum"?
I have tried 

```
select * from st_read('simple.gdb lorermIpsum') limit 10;
```

but it does not work.

Thank you

blackrez:
It is related to https://github.com/duckdblabs/duckdb_spatial/issues/36.

stev-0:
It looks to me like it's implemented in the code which calls GDAL:
https://github.com/duckdblabs/duckdb_spatial/blob/main/spatial/src/spatial/gdal/functions/st_read.cpp#L169-L197

but not yet implemented in the duckdb binding code - there is no `layer` option:

https://github.com/duckdblabs/duckdb_spatial/blob/main/spatial/src/spatial/gdal/functions/st_read.cpp#L411

But I'm pretty new to the code so could be wrong.

stev-0:
> It is related to #36.

Not sure it is, in GDAL layers within files are specified as part of the path to the data source, not as an open option.

blackrez:
> > It is related to #36.
> 
> Not sure it is, in GDAL layers within files are specified as part of the path to the data source, not as an open option.

Read this it describe more specially why it reads the first layer : https://github.com/duckdblabs/duckdb_spatial/issues/36#issuecomment-1500277740

Maxxen:
Its an issue on our end, we always open the first layer since we don't have a parameter for the requested layer yet, even though the rest of the code is ready for it.  Ill make a fix for this as soon as I get back on Monday.

aborruso:
It's very important to add support to layer, because a lot of format are dataset, with inside one or more resources: GPKG, FileGDB, OpenFileGDB, ecc..

Maxxen:
As soon as CI passed in #47 I'll merge it.

aborruso:
> As soon as CI passed in #47 I'll merge it.

What will be the syntax to read only the record of one layer?

Will there also be the possibility to read only the list of layers with their geometric typology?

Maxxen:
You can now pass a `st_read('...', layer=<layer_name>)` option to specify which layer.
You can also pass an integer to specify layer by index.
[...]

PR #775, opened 2026-03-18
fix(gdal): st_read_meta returns empty results for /vsi* paths

jatorre:
## Summary

`st_read_meta()` silently returns 0 rows when called with GDAL virtual filesystem paths (`/vsizip/`, `/vsicurl/`, `/vsigs/`, etc.). This is because the `Bind` function uses `MultiFileReader::CreateFileList()` which doesn't understand `/vsi*` prefixes and returns an empty list (masked by `ALLOW_EMPTY`).

This is inconsistent with `st_read()`, which correctly handles `/vsi*` paths by passing them directly to GDAL via `GetPrefix()`.

## The Fix

When the input path starts with `/vsi`, bypass `MultiFileReader` and pass the path directly to GDAL — matching the existing behavior of `st_read()`. Non-`/vsi` paths continue to use `MultiFileReader` for glob expansion.

**Before:**
```sql
-- Returns 0 rows silently
SELECT * FROM st_read_meta('/vsizip//vsicurl/https://example.com/data.zip');
```

**After:**
```sql
-- Returns full metadata (layers, geometry types, CRS, fields)
SELECT * FROM st_read_meta('/vsizip//vsicurl/https://example.com/data.zip');
```

## Root Cause

In `gdal_module.cpp`, the `ST_Read_Meta::Bind` function:

```cpp
const auto mfreader = MultiFileReader::Create(input.table_function);
const auto mflist = mfreader->CreateFileList(context, input.inputs[0], FileGlobOptions::ALLOW_EMPTY);
return make_uniq_base<FunctionData, BindData>(mflist->GetAllFiles());
```

`MultiFileReader::CreateFileList` tries to resolve `/vsizip/...` through DuckDB's VFS layer, fails, and returns an empty list. The `ALLOW_EMPTY` flag suppresses the error.

Meanwhile, `st_read` avoids this by using `GDALClientContextState::GetPrefix()` which already has explicit `/vsi` handling (line 446):

```cpp
if (StringUtil::StartsWith(value, "/vsi")) {
    return value;  // Pass through to GDAL unchanged
}
```

## Use Case

This enables listing layers in remote archives without downloading them:

```sql
-- Discover layers in a remote shapefile zip
SELECT unnest(layers) FROM st_read_meta('/vsizip//vsicurl/https://storage.googleapis.com/bucket/data.zip');

-- Works with signed URLs using the {braces} syntax
SELECT * FROM st_read_meta('/vsizip/{/vsicurl/https://...?X-Goog-Signature=...}/layer.shp');
```

Previously the only workaround was calling `ogrinfo` as an external process, or downloading the file locally first.

## Related Issues

- #45 — Original request for metadata reading (led to `st_read_meta`)
- #729 — Reports `st_read_meta` crashes/returns 0 rows for some formats
- #544 — `st_read_meta()` crash reports

### DIFF
--- a/src/spatial/modules/gdal/gdal_module.cpp
+++ b/src/spatial/modules/gdal/gdal_module.cpp
@@ -1172,6 +1172,17 @@ struct ST_Read_Meta {
 		// TODO: Add metadata, domains, relationships
 
 		// Get the filename list
+		const auto file_name = input.inputs[0].GetValue<string>();
+
+		// If the path starts with /vsi (GDAL virtual filesystem), pass it directly to GDAL.
+		// MultiFileReader doesn't understand /vsizip/, /vsicurl/, etc. and silently returns
+		// an empty file list, causing st_read_meta to return 0 rows.
+		if (StringUtil::StartsWith(file_name, "/vsi")) {
+			vector<OpenFileInfo> files;
+			files.emplace_back(file_name);
+			return make_uniq_base<FunctionData, BindData>(std::move(files));
+		}
+
 		const auto mfreader = MultiFileReader::Create(input.table_function);
 		const auto mflist = mfreader->CreateFileList(context, input.inputs[0], FileGlobOptions::ALLOW_EMPTY);
 		return make_uniq_base<FunctionData, BindData>(mflist->GetAllFiles());


src/spatial/modules/gdal/gdal_module.cpp
```diff
@@ -1172,6 +1172,17 @@ struct ST_Read_Meta {
 		// TODO: Add metadata, domains, relationships
 
 		// Get the filename list
+		const auto file_name = input.inputs[0].GetValue<string>();
```
Maxxen, line 1175: I don't think this will work. One of the things the MultiFileReader abstraction does is add an overload to the table function that also takes VARCHAR[] as the first input. If you pass that here and then try to call GetValue<string>() you will get the string representation of the input list, not an actual path.

jatorre:
Closing in favor of #779, which includes this fix and also adds support for GDAL driver-prefixed URLs (WFS, OAPIF, etc.).

PR #776, opened 2026-03-18
fix(gdal): set traditional GIS axis order when writing with SRS

jatorre:
## Summary

COPY TO with `SRS 'EPSG:4326'` fails for KML (and potentially other drivers that validate coordinates) because GDAL 3.x defaults to authority-compliant axis order (lat, lon) for EPSG:4326, while DuckDB GEOMETRY always stores coordinates in traditional GIS order (lon, lat).

**Before:**
```sql
COPY data TO 'output.kml' WITH (FORMAT GDAL, DRIVER 'KML', SRS 'EPSG:4326');
-- IO Error: Latitude 101.446569 is invalid. Valid range is [-90,90]
```

**After:**
```sql
COPY data TO 'output.kml' WITH (FORMAT GDAL, DRIVER 'KML', SRS 'EPSG:4326');
-- Works correctly
```

## Root Cause

In GDAL 3.x, `OSRSetFromUserInput("EPSG:4326")` creates a spatial reference with authority-compliant axis order — latitude first, longitude second. When GDAL's KML driver receives a geometry with coordinates in (lon, lat) order (as DuckDB always stores them), it interprets the first value as latitude and rejects any longitude > 90° as an invalid latitude.

## The Fix

One line: call `OSRSetAxisMappingStrategy(srs, OAMS_TRADITIONAL_GIS_ORDER)` after setting the SRS from user input. This tells GDAL to expect coordinates in (lon, lat) order, matching DuckDB's internal representation.

This is consistent with how GDAL's own documentation recommends handling the axis order change introduced in GDAL 3.0: https://gdal.org/tutorials/osr_api_tut.html#crs-and-axis-order

## Impact

Affects all COPY TO operations with `SRS` option where the target driver validates coordinate ranges. KML is the most common case, but any driver that checks axis-order-sensitive coordinate bounds would be affected.

The fix has no effect on drivers that don't validate coordinates (like Shapefile, GeoJSON) — they already work correctly because they write coordinates as-is regardless of axis order.

### DIFF
--- a/src/spatial/modules/gdal/gdal_module.cpp
+++ b/src/spatial/modules/gdal/gdal_module.cpp
@@ -1303,14 +1303,6 @@ auto Bind(ClientContext &context, CopyFunctionBindInput &input, const vector<str
 			continue;
 		}
 
-		if (StringUtil::CIEquals("ALWAYS_XY", option.first)) {
-			if (option.second.size() != 1) {
-				throw BinderException("GDAL COPY option 'ALWAYS_XY' only accepts a single value");
-			}
-			result->always_xy = option.second.back().GetValue<bool>();
-			continue;
-		}
-
 		throw BinderException("Unknown GDAL COPY option: '%s'", option.first);
 	}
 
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks! Could you retarget this to the v1.5 branch?

Maxxen:
I guess this should also be controlled by the "geometry_always_xy" option to not silently break existing workflows.

jatorre:
Thanks @Maxxen! Addressed both points:

1. **Retargeted to `v1.5-variegata`** — rebased on top of the current branch (resolved conflict with the new SRS resolution logic).

2. **Added `ALWAYS_XY` COPY option** — the axis mapping is now controlled by a boolean option that defaults to `true` (since DuckDB GEOMETRY is always lon,lat). Users can opt out with `ALWAYS_XY false` if needed:

```sql
-- Default: traditional GIS order (lon, lat) — fixes KML and similar drivers
COPY t TO 'out.kml' WITH (FORMAT GDAL, DRIVER 'KML', SRS 'EPSG:4326')

-- Opt out: authority-compliant axis order
COPY t TO 'out.kml' WITH (FORMAT GDAL, DRIVER 'KML', SRS 'EPSG:4326', ALWAYS_XY false)
```

Naming follows the existing `always_xy` parameter in `ST_Transform`.

Maxxen:
I don't think it should be a COPY option, there is a global configure option to control this behavior for all cases: https://duckdb.org/2026/03/09/announcing-duckdb-150#breaking-change-flipping-of-axis-order

jatorre:
YEp! Sorry, too much info.

Updated. removed the ALWAYS_XY COPY option and now reading the global geometry_always_xy setting via SpatialSettings::AlwaysXY() instead. 

Also added a warning when the setting isn't configured, same pattern as ST_Transform and the spheroid functions. Let me know if you think is appropiate.

So SET geometry_always_xy = true + COPY TO ... WITH (FORMAT GDAL, DRIVER 'KML', SRS 'EPSG:4326') just works now.

PR #779, opened 2026-03-24
fix: support GDAL driver-prefixed URLs (WFS, OAPIF, etc.)

jatorre:
## Summary

- GDAL driver-prefixed URLs like `WFS:https://...` and `OAPIF:https://...` were being mangled by `AddPrefix()` prepending `/vsiduckdb-<UUID>/`, making them unrecognizable to `GDALOpenEx`
- Added detection for the `DRIVERNAME:http(s)://` pattern (all-uppercase driver name followed by an HTTP URL) and passes these through directly, matching the existing behavior for `/vsi*` paths
- Also fixes `st_read_meta` to bypass `MultiFileReader` for these URLs and `/vsi*` paths, so they reach GDAL properly instead of being resolved as local file globs
- Security: driver-prefixed URLs are blocked when `enable_external_access=false`, same as `/vsi*` paths

This enables queries like:
```sql
SELECT * FROM st_read('WFS:https://ahocevar.com/geoserver/wfs', layer='topp:states');
SELECT * FROM st_read('OAPIF:https://demo.pygeoapi.io/master', layer='lakes');
SELECT * FROM st_read_meta('WFS:https://ahocevar.com/geoserver/wfs');
```

## Test plan
- [x] Added tests for `st_read` blocking WFS/OAPIF URLs when external access is disabled
- [x] Added test for `st_read_meta` blocking driver-prefixed URLs when external access is disabled
- [x] All 178 existing GDAL test assertions pass
- [x] Manually verified `st_read` with live WFS and OAPIF endpoints
- [x] Manually verified `st_read_meta` with live WFS endpoint

### DIFF
--- a/src/spatial/modules/gdal/gdal_module.cpp
+++ b/src/spatial/modules/gdal/gdal_module.cpp
@@ -429,11 +429,39 @@ public:
 		VSIFileManager::RemoveHandler(client_prefix);
 	}
 
+	// Check if a path looks like a GDAL driver-prefixed URL (e.g., "WFS:https://...", "OAPIF:https://...")
+	// These need to be passed directly to GDALOpenEx without our custom VSI prefix.
+	static bool IsGDALDriverPrefixedURL(const string &value) {
+		auto colon_pos = value.find(':');
+		if (colon_pos == string::npos || colon_pos == 0 || colon_pos > 20) {
+			return false;
+		}
+		// Check that the prefix is all uppercase letters (GDAL driver names are uppercase)
+		for (idx_t i = 0; i < colon_pos; i++) {
+			char c = value[i];
+			if (!StringUtil::CharacterIsAlpha(c) || c != StringUtil::CharacterToUpper(c)) {
+				return false;
+			}
+		}
+		// Check that after the colon there is a URL scheme (http:// or https://)
+		// This excludes database connection strings like "PG:dbname=..." which are not supported.
+		auto rest = value.substr(colon_pos + 1);
+		return StringUtil::StartsWith(rest, "http://") || StringUtil::StartsWith(rest, "https://");
+	}
+
 	string AddPrefix(const string &value) const {
 		// If the user explicitly asked for a VSI prefix, we don't add our own
 		if (StringUtil::StartsWith(value, "/vsi")) {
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks! Does this supersede https://github.com/duckdb/duckdb-spatial/pull/775 or is that unrelated?

src/spatial/modules/gdal/gdal_module.cpp
```diff
@@ -1774,11 +1802,20 @@ auto Bind(ClientContext &context, TableFunctionBindInput &input, vector<LogicalT
...
 
-	const auto mf_reader = MultiFileReader::Create(input.table_function);
-	const auto mf_inputs = mf_reader->CreateFileList(context, input.inputs[0], FileGlobOptions::ALLOW_EMPTY);
-
 	auto result = make_uniq<BindData>();
-	result->files = mf_inputs->GetAllFiles();
+
+	// For /vsi* paths and GDAL driver-prefixed URLs (e.g., "WFS:https://..."),
+	// bypass MultiFileReader since these are not local file globs.
+	auto raw_path = input.inputs[0].GetValue<string>();
```
Maxxen: I don't think this will work. One of the things the `MultiFileReader` abstraction does is add an overload to the table function that also takes `VARCHAR[]` as the first input. If you pass that here and then try to call `GetValue<string>()` you will get the string representation of the input list, not an actual path.

jatorre:
Yes, this supersedes #775 — it includes the same `/vsi*` fix for `st_read_meta` plus adds support for GDAL driver-prefixed URLs (`WFS:`, `OAPIF:`, etc.) in both `st_read` and `st_read_meta`.

src/spatial/modules/gdal/gdal_module.cpp
jatorre: Good catch! Fixed in 254588f — now we check `input_val.type().id() == LogicalTypeId::VARCHAR` before calling `GetValue<string>()`. When the input is a `LIST` (from the `VARCHAR[]` overload), we fall through to `MultiFileReader` as before.

jatorre:
Sorry, found the issues separately and did not connect them after. thanks.

Maxxen:
Thanks, LGTM

PR #791, opened 2026-04-11
fix: only suppress FID column when it duplicates an existing attribute field

### LINKED ISSUE #790: ST_Read still suppresses FID for GeoPackage files (INCLUDE_FID=NO bug)
## What happens

`ST_Read` does not expose the `fid` column when reading GeoPackage files on DuckDB v1.5.1 with the latest spatial extension (`a415e74`).

This was supposed to be fixed by #713 (which closed #213), but the read path in `gdal_module.cpp` still contains logic that suppresses the FID for any format where GDAL reports a named FID column, which includes GeoPackage.

## Root cause

[`gdal_module.cpp` L614–618](https://github.com/duckdb/duckdb-spatial/blob/a415e74/src/spatial/modules/gdal/gdal_module.cpp#L614-L618):

```cpp
// Check FID column
const auto fid_col = OGR_L_GetFIDColumn(layer);
if (fid_col && strcmp(fid_col, "") != 0) {
    // Do not include the explicit FID if we already have it as a column
    result->layer_options.AddString("INCLUDE_FID=NO");
}
```

The comment assumes the named FID column already appears as a regular attribute in the Arrow stream. That's not the case, GDAL classifies the GeoPackage `fid` as the layer FID, not a regular field. Setting `INCLUDE_FID=NO` then suppresses it entirely, so neither `fid` nor `OGC_FID` appears in the output.

The OGC_FID feature works only for formats where `OGR_L_GetFIDColumn()` returns an empty string (synthetic/unnamed FID), which is the opposite of the common case.

## Reproduction

```python
import duckdb, tempfile, os

conn = duckdb.connect()
conn.install_extension("spatial")
conn.load_extension("spatial")

with tempfile.TemporaryDirectory() as tmp:
    gpkg = os.path.join(tmp, "test.gpkg")
    conn.execute(f"""
        COPY (
            SELECT 'A' AS name,
                   ST_GeomFromText('POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))') AS geom
            UNION ALL
            SELECT 'B',
                   ST_GeomFromText('POLYGON((2 2, 3 2, 3 3, 2 3, 2 2))')
        ) TO '{gpkg}' WITH (FORMAT GDAL, DRIVER 'GPKG')
    """)

    cols = [r[0] for r in conn.execute(
        f"DESCRIBE SELECT * FROM st_read('{gpkg}')"
    ).fetchall()]
    print(cols)  # ['name', 'geom'] — no fid / OGC_FID
```

## Expected

The output should include a `fid` (or `OGC_FID`) column with values `[1, 2]`. The GeoPackage file does contain these values (verifiable via `sqlite3`).

## Suggested fix

Remove the `INCLUDE_FID=NO` block. Leaving `INCLUDE_FID` at its default (`YES`) causes GDAL to emit the FID as a column. No duplication occurs for GeoPackage, GDAL does not expose the FID as a regular OGR field.

## Environment

- DuckDB: v1.5.1
- Spatial extension: `a415e74`
- OS: macOS (Apple Silicon)

rernst76:
## Summary
- Only set `INCLUDE_FID=NO` when the FID column name already exists as a regular attribute field (`OGR_FD_GetFieldIndex`), preventing duplicate columns for formats like GeoJSON
- For formats where the FID is separate (GeoPackage, GDB, etc.), the FID column now correctly appears in output

Closes #790

## Test plan
- [x] New `st_read_gpkg_fid.test` — round-trips a GeoPackage and asserts `fid` column is present with correct values
- [x] Updated `st_read_gdb.test` — expects `OBJECTID` FID column in output
- [x] All 11 GDAL tests pass, including existing GeoJSON tests (no duplication)

### DIFF
--- a/src/spatial/modules/gdal/gdal_module.cpp
+++ b/src/spatial/modules/gdal/gdal_module.cpp
@@ -646,12 +646,15 @@ auto Bind(ClientContext &ctx, TableFunctionBindInput &input, vector<LogicalType>
 		// Get the layer geometry type if available
 		result->layer_type = OGR_L_GetGeomType(layer);
 
-		// Check FID column
+		// Only suppress the FID if the layer already exposes it as a regular attribute field
 		const auto fid_col = OGR_L_GetFIDColumn(layer);
 		if (fid_col && strcmp(fid_col, "") != 0) {
-			// Do not include the explicit FID if we already have it as a column
-			result->layer_options.AddString("INCLUDE_FID=NO");
+			const auto layer_defn = OGR_L_GetLayerDefn(layer);
+			if (OGR_FD_GetFieldIndex(layer_defn, fid_col) >= 0) {
+				result->layer_options.AddString("INCLUDE_FID=NO");
+			}
 		}
+
 		const auto geom_col_name = OGR_L_GetGeometryColumn(layer);
 
 		// Get the arrow stream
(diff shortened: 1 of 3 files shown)


Maxxen:
Thanks!

PR #827, opened 2026-06-15
Add spatial join bbox join-filter-pushdown

Maxxen:
...and fix ST_DistanceWithin const-serialization.

We can make the join-filter-pushdown much more advanced by pushing down the entire r-tree, but this is good enough for now and allows us to utilize stats pruning.

CC @lnkuiper, im always nervous about column binding/storage index stuff in the optimizer, does it look okay?

### DIFF
--- a/src/spatial/modules/main/spatial_functions_scalar.cpp
+++ b/src/spatial/modules/main/spatial_functions_scalar.cpp
@@ -16,6 +16,8 @@
 #include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/common/vector_operations/septenary_executor.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/common/serializer/serializer.hpp"
+#include "duckdb/common/serializer/deserializer.hpp"
 
 #include "spatial/util/distance_extract.hpp"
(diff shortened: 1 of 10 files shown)


lnkuiper:
Looks great!

There is one small issue with recursive CTEs that should be fixed:
```sql
# name: test/temp.test
# description: Reproducer for stale spatial join dynamic filters in recursive CTEs
# group: [temp]

require spatial

statement ok
CREATE TABLE probe AS
SELECT x, ST_Point(x, 0) AS geom
FROM (VALUES (0), (100), (200)) v(x);

# The build-side bounding box changes each recursive iteration. Dynamic filters
# pushed by the spatial join must be cleared before the next execution of the
# same physical operator, otherwise the probe scan is filtered by stale bounds.
query I
WITH RECURSIVE r(i, geom) AS (
	SELECT 0, ST_Point(0, 0)
	UNION ALL
	SELECT r.i + 1, p.geom
	FROM probe p, r
	WHERE r.i < 2
	  AND ST_Intersects(p.geom, ST_Point((r.i + 1) * 100, 0))
)
SELECT count(*) FROM r;
----
3

```
See `PhysicalHashJoin` for the fix: `ClearFilters(op)`.

PR #832, opened 2026-06-23
Add relassert and re-run tests to stabilize CI

smvv:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9787

I want to repeatedly run extensions tests to detect (more) ASAN failures in CI. Some tests can fail "randomly" (if the right code paths are executed) and that affects duckdb CI on main and in PRs (and some customers that build duckdb from main).

The extension author is not always aware of failure when bumping the extension in duckdb, and re-running tests should reduce the chance of merging code that fails once every ~ten runs.

I'm trying the new CLI flag `--stabilize-tests` in this extension first to see what we discover. It could be that we move the changes elsewhere (like `extension-ci-tools`) in the future.


Maxxen:
So main branch isn't really up-to-date yet (patches not applied). Not sure if we want to run this on v1.5 instead?

Issue #824, opened 2026-06-01
Can't create RTREE spatial index using new CRS-scoped GEOMETRY column type


mhkeller:
When trying to create a RTREE spatial index with a geometry field that has a CRS, I get an error that I can only do this on plain `GEOMETRY` columns. Here's a reproduction ([view example csv](https://gist.github.com/mhkeller/4dba2d7cff419aa64d622727c91ee3b8))

```sql
INSTALL spatial;
LOAD spatial;

.print '=== DuckDB version (need v1.5.0 or newer) ==='
SELECT version() AS duckdb_version;

.print '=== 1. Control: plain GEOMETRY + RTREE ==='

CREATE OR REPLACE TABLE lighthouses_plain AS
  SELECT
    name,
    country,
    longitude,
    latitude,
    height_m,
    year_built,
    ST_Point(longitude, latitude)::GEOMETRY AS geom
  FROM read_csv_auto('lighthouses.csv');

CREATE INDEX IF NOT EXISTS lighthouses_plain_geom_idx
  ON lighthouses_plain USING RTREE (geom);

.print '=== 2. Bug repro: GEOMETRY(''OGC:CRS84'') + RTREE ==='

CREATE OR REPLACE TABLE lighthouses_crs AS
  SELECT
    name,
    country,
    longitude,
    latitude,
    height_m,
    year_built,
    ST_Point(longitude, latitude)::GEOMETRY('OGC:CRS84') AS geom
  FROM read_csv_auto('lighthouses.csv');

-- Binder Error: RTree indexes can only be created over GEOMETRY columns.
CREATE INDEX IF NOT EXISTS lighthouses_crs_geom_idx
  ON lighthouses_crs USING RTREE (geom);

```

Maxxen:
Hey! Thanks for reporting this issue!
This seems like an oversight indeed and should be a relatively easy fix.

mhkeller:
Just wanted to check in on what you think a timeline might be for this. I'm coordinating a dependency update (has a few moving parts so I try to space them out) and want to see if I should do it now or wait a little longer. Thanks again!

Maxxen:
Sorry about that, got lost in the weeds. Ill have a fix ready for v1.5.5

mhkeller:
Thanks!

PR #839, opened 2026-07-08
Support geometry with CRS in R-Tree columns, Improve R-Tree scans

Maxxen:
In this PR we:
- Add support for creating R-Tree indexes on geometry columns with CRS, closing #824
- Fix a de/serialization issue related to virtual columns in r-tree scans
- Make the data-fetching part of the R-Tree index scan multi-threaded (the actual index traversal is still behind a lock)

The last point improves the performance of index scans by about 50% on my synthetic benchmarks, but I would assume the real gain is bigger on wider tables and thicker columns.

**25M random points test**
```
load spatial;
.mode trash

.timer on
.print 'Create data'
CREATE TABLE t1 AS SELECT row_number() OVER () AS id, point::GEOMETRY AS geom
                   FROM st_generatepoints({min_x: 0, min_y: 0, max_x: 1000, max_y: 1000}::BOX_2D, 25_000_000, 1337);

.print 'Create index'

.print 'Begin scan'
SELECT count(*) FROM t1 WHERE ST_Within(geom, ST_MakeEnvelope(100, 100, 700, 700));
```

**Current DuckDB v1.5.4:**
```
Create data
Run Time (s): real 1.083 user 1.030571 sys 0.052126
Create index
Run Time (s): real 3.202 user 5.849422 sys 0.269274
Begin scan
Run Time (s): real 8.305 user 8.230279 sys 0.060562
```

**This PR:**
```
Create data
Run Time (s): real 1.051 user 1.001588 sys 0.049487
Create index
Run Time (s): real 3.121 user 5.680956 sys 0.256225
Begin scan
Run Time (s): real 4.801 user 13.107156 sys 24.429724
```

### DIFF
--- a/src/spatial/index/rtree/rtree_index_create_logical.cpp
+++ b/src/spatial/index/rtree/rtree_index_create_logical.cpp
@@ -72,16 +72,17 @@ static PhysicalOperator &CreateNullFilter(PhysicalPlanGenerator &generator, cons
 }
 
 static PhysicalOperator &CreateBoundingBoxProjection(PhysicalPlanGenerator &planner, const LogicalOperator &op,
-                                                     const vector<LogicalType> &types, ClientContext &context) {
+                                                     const vector<LogicalType> &types, const LogicalType &geom_type,
+                                                     ClientContext &context) {
 	auto &catalog = Catalog::GetSystemCatalog(context);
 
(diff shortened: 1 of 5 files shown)


mhkeller:
Thanks again!

Issue #744, opened 2026-01-18
ST_Read fails on non-UTF-8 Shapefile (macOS only?)


yutannihilation:
I found `ENCODING=` GDAL option doesn't work. I don't remember if this used to work on macOS, but at least this works on Windows. So, maybe something is wrong with macOS build?

```sql
FROM ST_Read('/path/to/data.shp', open_options = ["ENCODING=CP932"])
```

The above doesn't fail and silently returns the results like below.

```
┌────────────┬────────────┬────────────┐
│ A22_000001 │ A22_000002 │ A22_000003 │
│  varchar   │  varchar   │  varchar   │
├────────────┼────────────┼────────────┤
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
│ ????INVA…  │ ????INVA…  │ ????INVA…  │
├────────────┴────────────┴────────────┤
│ 10 rows                    3 columns │
└──────────────────────────────────────┘
```

Also, `COPY TO` fails if we specify the option. 

```sql
COPY (
  SELECT ST_Point(1, 1), 'あああ' as text
) TO 'tmp'
WITH (FORMAT GDAL, DRIVER 'ESRI Shapefile', LAYER_CREATION_OPTIONS 'ENCODING=CP932');
```
```
IO Error:
GDAL Error (1): Failed to create field name 'text': cannot convert to CP932
```

Note that, `COPY TO`'s case isn't serious as we rarely output non-UTF-8 files. However, I really wish `ST_Read` works because, unfortunately, there are a lot of non-UTF-8 geospatial data.

yutannihilation:
With DuckDB v1.4.4, `ST_Read()` fails rather than shows us invalid strings. I hope the stack trace might help.

```
D FROM ST_Read('path/to/data.shp', open_options = ["ENCODING=CP932"]) LIMIT 10;
[...]

Issue #394, opened 2024-09-10
Encoding issue with geospatial data (shapefile)


vlebert:
### What happens?

When trying to import a shapefile encoded with CP1252 I have the following error 

`InvalidInputException: Invalid Input Error: Invalid unicode (byte sequence mismatch) detected in segment statistics update
`

I tried various options in the st_read but no success

A current workaround is to convert first shapefile to geoparquet with ogr2ogr and then import the geoparquet

### To Reproduce

```
CREATE TABLE t_adresse AS SELECT * FROM st_read('source/t_adresse.shp');
```

Note : the shapefile does have a `.cpg` file providing the encoding

Even forcing encoding do fail:

```
CREATE TABLE t_adresse AS SELECT * FROM st_read('source/t_adresse.shp', open_options=['ENCODING=WINDOWS-1252']);
CREATE TABLE t_adresse AS SELECT * FROM st_read('source/t_adresse.shp', open_options=['ENCODING=CP-1252']);
CREATE TABLE t_adresse AS SELECT * FROM st_read('source/t_adresse.shp', open_options=['ENCODING=CP1252']);

```

However, `SELECT * FROM ST_READ('source/t_adresse.shp')` does not gives error (in python)

Current workaround :

```
ogr2ogr -f parquet t_adresse.parquet source/t_adresse.shp

CREATE TABLE t_adresse AS SELECT * from 't_adresse.parquet'

```

### OS:

MacOS

### DuckDB Version:

1.1

### DuckDB Client:

Python

### Full Name:

Valérian LEBERT

### Affiliation:

Digi-Studio

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

No - I cannot easily share my data sets due to their large size

### Did you include all code required to reproduce the issue?

- [X] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [X] Yes, I have

vlebert:
Using the experimental `st_readshp` (without extra argument), I could load the dataset. All text fields are loaded as `blob`

rouault:
> i think the issue is that we dont bundle the (optional) iconv library that gives GDAL the capability to re-encode text.

https://github.com/OSGeo/gdal/pull/10799 should improve that situation

casperhart:
[...]

Issue #356, opened 2024-06-30
SHP driver open_options does not seem to work with ENCODING option


ttomasz:
Example files:
- small - https://opendata.geoportal.gov.pl/prg/adresy/PunktyAdresowe/02/020101.zip
- large - https://integracja.gugik.gov.pl/PRG/pobierz.php?adresy_zbiorcze_shp

SQL example:
```sql
select *
from st_read('/vsizip/./Downloads/020101.zip/PRG_PunktyAdresowe_020101.shp', open_options=['ENCODING=WINDOWS-1250']) 
limit 5;
```

Result:
```
┌─────────┬─────────┬─────────┬─────────────┬─────────┬─────────────────────┬─────────┬───────────────────────────────────────────────┐
│  TERYT  │   PNA   │ SIMC_id │ SIMC_nazwa  │ ULIC_id │     ULIC_nazwa      │  Numer  │                     geom                      │
│ varchar │ varchar │ varchar │   varchar   │ varchar │       varchar       │ varchar │                   geometry                    │
├─────────┼─────────┼─────────┼─────────────┼─────────┼─────────────────────┼─────────┼───────────────────────────────────────────────┤
│ 020101  │ 59-700  │ 0935989 │ ????INVAL…  │ 18648   │ ????INVALID VALUE…  │ 34      │ POINT (260060.88199839657 382544.5259835962)  │
│ 020101  │ 59-700  │ 0935989 │ ????INVAL…  │ 09421   │ ????INVALID VALUE…  │ 82      │ POINT (262141.78239842618 381462.537184705)   │
│ 020101  │ 59-700  │ 0935989 │ ????INVAL…  │ 02058   │ ????INVALID VALUE…  │ 13      │ POINT (258755.64489838033 383519.40798285604) │
│ 020101  │ 59-700  │ 0935989 │ ????INVAL…  │ 07026   │ ????INVALID VALUE…  │ 27C     │ POINT (261791.14839842354 382338.35378450714) │
│ 020101  │ 59-700  │ 0935989 │ ????INVAL…  │ 07026   │ ????INVALID VALUE…  │ 27B     │ POINT (261788.7517984234 382329.84118450433)  │
└─────────┴─────────┴─────────┴─────────────┴─────────┴─────────────────────┴─────────┴───────────────────────────────────────────────┘
```

Using that encoding in open options works with ogrinfo.

rouault:
> Using that encoding in open options works with ogrinfo.

[...]

PR #844, opened 2026-07-12
Upgrade GDAL to 3.13.1 (adds iconv-free  ISO-8859-2 and -15 and CP437/CP1250/CP1251/CP1252 support)

casperhart:
Related to #394 #356 and possibly #744

## Motivation

Reading DBF/shapefile attributes encoded in **CP1252** (and related single-byte
encodings) currently produces errors. This is because the vendored GDAL
overlay port is pinned to **3.8.5**, which predates
[OSGeo/gdal#10799](https://github.com/OSGeo/gdal/pull/10799) — a change that
makes CP1252→UTF-8 recoding always available even in a minimal GDAL build that
lacks `iconv`, like what duckdb-spatial uses. That fix landed in GDAL **3.10**.

This PR bumps the overlay port to the latest minor release, **GDAL 3.13.1**, and
adapts the extension to the GDAL API changes that accumulated since 3.8.5.

So now queries like this which previously produced errors, can now return a result:
```
memory D from st_read('MASTER.DBF', open_options=>['ENCODING=CP1252']) where isbn='672769';
INTERNAL Error: invalid code point detected in Utf8Proc::UTF8ToCodepoint, likely due to invalid UTF-8
```

Now:
```
memory D select title from st_read('MASTER.DBF', open_options=>['ENCODING=CP1252']) where isbn='672769';
┌───────────────────────────────────────────────────────────────────────────┐
│                                   TITLE                                   │
│                                  varchar                                  │
├───────────────────────────────────────────────────────────────────────────┤
│ CHRIS ACHILLEOS – HOST OF CHAOS MINI POSTER 40X50CM  (RARE SOME WEAR) 535 │
└───────────────────────────────────────────────────────────────────────────┘
```

## Changes

Bump the vendored GDAL overlay port from 3.8.5 to 3.13.1 so that DBF/ shapefile attributes in CP1252 (and related single-byte encodings) can be recoded to UTF-8 without an iconv dependency (GDAL PR #10799, landed in 3.10).

vcpkg port:
- Bump version + SHA512, refresh find-link-libraries/target-is-valid from current upstream vcpkg (the old copies no longer applied).
- Regenerate duckdb_gdal_json.patch (json_object_get -> gdal__json_object_get symbol rename) against the moved 3.13.1 sources.
- Drop duckdb_gdal_msys / remove_filehandler / windows_static patches: all are now upstream in GDAL (VSIRemovePluginHandler, calloc/stat64 fixes) or target code that was removed.

Extension:
- Adapt the custom VSI filesystem handler in gdal_module.cpp to GDAL's newer VSIVirtualHandle / VSIFilesystemHandler API: 2-arg Read/Write (3.13), new Error()/ClearErr() virtuals (3.10), Open() returning VSIVirtualHandleUniquePtr, const IsLocal(), and Rename() with progress args.


casperhart:
I just noticed #838 merged in to main a few days ago, which resolves this already, so I will close this. Looking forward to the new version, thank you!

PR #845, opened 2026-07-12
Add optional parameters to ST_MakeValid

goulartdev:
This PR add support to two optional parameters in `ST_MakeValid`:

`method`: accepts `LINEWORK` (the default) or `STRUCTURE`.

- The `LINEWORK` method combines all rings into a set of noded lines and then extracts valid polygons from that linework. This method keeps all input vertices.
- The `STRUCTURE` method first makes all rings valid then merges shells and subtracts holes from shells to generate valid result. It assumes that holes and shells are correctly categorized.

`keepCollapsed`: whether or not to retain components that have collapsed into a lower dimensionality. Only works with the  `STRUCTURE` method. The default value is `true`.

These parameters were added to GEOS in version 3.10.


Maxxen:
Thanks!

PR #853, opened 2026-08-01
Fix optimizer misidentifying non-geometry && as spatial join predicate

### LINKED ISSUE #852: `&&` operator incorrectly rewritten into spatial join for non-geometry `LIST` columns, crashing with `ST_IsEmpty(UBIGINT[])` internal error
**Environment:**
DuckDB v1.5.4 (also repros on v1.5.2), spatial ext 28db190 (also dc1996b), macOS

**Summary:**
The spatial extension's optimizer pattern-matches `&&` purely on the operator token and rewrites any join using it into `PhysicalSpatialJoin`, whose local sink state unconditionally calls `ST_IsEmpty(<join-key-type>)` assuming `GEOMETRY` — without checking the actual bound type. Using `&&` between two plain `LIST` columns (e.g. `UBIGINT[]`) crashes with an internal error and invalidates the whole connection, requiring a restart.

**Repro (minimal, no geometry needed):**
````sql
create table a as select * from (values (1,[10,20,30]::ubigint[]),(2,[40,50]::ubigint[])) t(id,cells);
create table b as select * from (values (1,[20,25]::ubigint[]),(2,[99]::ubigint[])) t(id,cells);
select a.id, b.id from a join b on a.cells && b.cells;
````

→ `INTERNAL Error: Failed to find function ST_IsEmpty(UBIGINT[])`, then subsequent queries fail with `database has been invalidated`.

### PR DESCRIPTION
**Pull request by oglego**
Fixes #852 

The spatial extension's optimizer pattern-matches && purely on the operator token and rewrites any join using it into PhysicalSpatialJoin - the GDAL module (duckdb-spatial/src/spatial/modules/gdal/gdal_module.cpp) has a check for this:

```cpp
		if (func.children[0]->return_type.id() != LogicalTypeId::GEOMETRY ||
		    func.children[1]->return_type.id() != LogicalTypeId::GEOMETRY) {
			continue;
		}
```

I applied similar logic in spatial_join_optimizer.cpp (duckdb-spatial/src/spatial/operators/spatial_join_optimizer.cpp) and I believe that this resolves the bug found in #852 

I added a regression test that covers the original repro plus a sanity check that real GEOMETRY joins still get rewritten.

If I missed anything on this just let me know, thank you!!

### DIFF
--- a/src/spatial/operators/spatial_join_optimizer.cpp
+++ b/src/spatial/operators/spatial_join_optimizer.cpp
@@ -113,6 +113,12 @@ static bool IsSpatialJoinPredicate(const unique_ptr<Expression> &expr, const uno
 	if (spatial_predicate_map.count(func.function.name) == 0) {
 		return false;
 	}
+	
+	// The function's operands must be GEOMETRY 
+	if (func.children[0]->return_type.id() != LogicalTypeId::GEOMETRY ||
+	    func.children[1]->return_type.id() != LogicalTypeId::GEOMETRY) {
+		return false;
+	}
 
 	const auto left_side = JoinSide::GetJoinSide(*func.children[0], left_bindings, right_bindings);
 	const auto right_side = JoinSide::GetJoinSide(*func.children[1], left_bindings, right_bindings);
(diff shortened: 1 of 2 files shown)


Maxxen:
Thanks!

PR #855, opened 2026-08-06
Reset WKB dimension state between rows

matthiasgoergens:
No upstream issue exists — this defect was found via an adversarial review of the SGL trailing-data fix; this PR is the initial report.

The SGL WKB reader retained its mixed-Z/M flag between parses, even though
Spatial reuses one reader across every row in a vector. It also stopped
accumulating dimensions after the first child that differed from the root.
Together these behaviours could silently discard Z or M coordinates from a
mixed collection and from later rows.

Reset the mixed-dimension flag for every parse and accumulate Z/M presence from
the root and every descendant. This preserves the complete dimension union
when Spatial homogenises mixed WKB while keeping each row independent.

The SQL regression processes a mixed Z/M collection followed by an ordinary Z
point in one vector. A standalone SGL regression directly checks both the
dimension union and the reset state; it fails against the unfixed reader and
passes under Clang ASan/UBSan. The full Spatial `relassert` build and adjacent
WKB suites also pass.

---
Verified (2026-08-05): `test/sql/geometry/st_ashexwkb.test` fails on unpatched
duckdb-spatial main `2b072abd2a` (second row degrades to `POINT (1 2)`, mixed
collection keeps only Z) and passes with this branch (28 assertions).


matthiasgoergens:
The CI failure (Build extension binaries / linux_amd64) is a pre-existing build break between duckdb-spatial and duckdb core main, not caused by this PR. DuckDB core PR #24278 (merged 2026-08-05) changed `table_function_bind_t` to take `vector\<Identifier\>\&` instead of `vector\<string\>\&`. The extension's bind functions in spatial_functions_table.cpp, gdal_module.cpp, osm_module.cpp, and shapefile_module.cpp still use the old signature, so they fail to compile against current main.

The PR compiles and tests fine against the pinned submodule (2026-07-28). The CI workflow floats `duckdb_version: main`, which is what picks up the breaking change.

What's the preferred fix here — should I update the bind signatures in this PR, or is there a plan to pin `duckdb_version` in the workflow? Happy to go either way.

Maxxen:
Hello! Thanks for the PR 

Yeah this break is unrelated, if you want to fix it there's a bunch of patches in duckdb/.github/patches/extensions/spatial that need to be applied - but otherwise Il do it myself soon and rebase this PR after. 

You could also retarget to the `v1.5-variegata` branch which will feed into the next v1.5 (v1.5.6) release, given that this is a bugfix. But it can also stay on main - once the release is getting near I will probably backport some changes from main anyway.

matthiasgoergens:
> Thanks, and sorry for the slow reply. Retargeted this PR and #856 to `v1.5-variegata`: both rebased onto 83135b0 (the only conflicts were the new `sgl_test.cpp` test functions landing next to `test_linear_referencing`), relassert build, `st_ashexwkb.test` and the SGL unit tests pass there. CI should be green now that the branch has the submodule and patches from #873.

PR #857, opened 2026-08-15
fix: update winding number vertex handling in ST_Contains

### LINKED ISSUE #806: `ST_Contains(POLYGON_2D, POINT_2D)` false positives when point latitude equals polygon vertex latitude
# `ST_Contains(POLYGON_2D, POINT_2D)` false positive outside polygon bbox

`ST_Contains` returns `true` for a point that is clearly outside the polygon bbox when using the typed `POLYGON_2D` / `POINT_2D` overload. (Possibly related issue: #434)

## Repro

```sql
LOAD spatial;

WITH poly AS (
  SELECT ST_GeomFromText(
    'POLYGON ((
      11.6742158 48.156714,
      11.6781624 48.144285,
      11.6837275 48.1447925,
      11.6742158 48.156714
    ))'
  ) AS geom
),
pts AS (
  SELECT * FROM (VALUES
    (-123.1061120::DOUBLE, 48.144285::DOUBLE)
  ) AS t(lon, lat)
)
SELECT
  ST_XMin(geom) AS xmin,
  ST_XMax(geom) AS xmax,
  ST_Contains(geom::POLYGON_2D, ST_MakePoint(lon, lat)::POINT_2D) AS typed_contains,
  ST_Contains(geom, ST_MakePoint(lon, lat)::GEOMETRY) AS generic_contains
FROM poly, pts;
```

## Result

The polygon x-range is about `11.674–11.684`, but the test point has `lon = -123.106112.`

## Expected:

```
typed_contains = false
generic_contains = false
```

## Actual:

```
typed_contains = true
generic_contains = false
```

This seems to happen when the point latitude equals a local-minimum vertex latitude of the polygon.

## Environment

OS: Linux 6.19.14
DuckDB version: `v1.5.2`
Spatial Extension version:
```
┌───────────────────┬──────────────┬────────────────┐
│ extension_version │ install_mode │ installed_from │
│      varchar      │   varchar    │    varchar     │
├───────────────────┼──────────────┼────────────────┤
│ dc1996b           │ REPOSITORY   │ core           │
└───────────────────┴──────────────┴────────────────┘
```

oglego:
Fixes #806

I believe all that needs to be changed on this is in the inequality checks below:

Original:

```cpp
} else if (side == Side::LEFT && (y1 < y && y <= y2)) {
  winding_number++;
} else if (side == Side::RIGHT && (y2 <= y && y < y1)) {
  winding_number--;
}
```

This can be found in lines 2087-2091 (src/spatial/modules/main/spatial_functions_scalar.cpp) - the change is just in the consistency on the winding number loops endpoint handling:

Updated:

```cpp
} else if (side == Side::LEFT && (y1 <= y && y < y2)) {
  winding_number++;
} else if (side == Side::RIGHT && (y2 <= y && y < y1)) {
  winding_number--;
}
```

Set (y1 <= y && y < y2) instead of (y1 < y && y <= y2)

I dug around on how GEOS handled ray crossing and found their `RayCrossingCounter::countSegment` with the below comments and inequality checks:

```cpp
// To avoid double-counting shared vertices, we use the convention that
// - an upward edge includes its starting endpoint, and excludes its
//   final endpoint
// - a downward edge excludes its starting endpoint, and includes its
//   final endpoint

if((p1.y > point.y && p2.y <= point.y) || (p2.y > point.y && p1.y <= point.y)) {
    // orientation determines whether this is counted as a crossing
}
```

With changing the inequality this will now enforce the same rule - a shared vertex on the scanline is claimed by exactly one of its two adjacent edges.

I've added the statement from #806 in a test suite and this seems to resolve the problem, I also added a couple of other tests to make sure it still worked for other scenarios as well.

Let me know if I am off in my thinking on this, and thanks in advance for the review!

### DIFF
--- a/src/spatial/modules/main/spatial_functions_scalar.cpp
+++ b/src/spatial/modules/main/spatial_functions_scalar.cpp
@@ -2084,7 +2084,7 @@ struct ST_Contains {
 						// return Contains::ON_EDGE;
 						contains = false;
 						break;
-					} else if (side == Side::LEFT && (y1 < y && y <= y2)) {
+					} else if (side == Side::LEFT && (y1 <= y && y < y2)) {
 						winding_number++;
 					} else if (side == Side::RIGHT && (y2 <= y && y < y1)) {
 						winding_number--;
--- a/test/sql/geometry/st_contains.test
+++ b/test/sql/geometry/st_contains.test
@@ -1,0 +1,85 @@
+# name: test/sql/geometry/st_contains.test
+# group: [geometry]
+#
+# https://github.com/duckdb/duckdb-spatial/issues/806
+
+require spatial
+
+# Query from example in issue 806
+
+query IIII
+WITH poly AS (
+  SELECT ST_GeomFromText(
+    'POLYGON ((11.6742158 48.156714, 11.6781624 48.144285, 11.6837275 48.1447925, 11.6742158 48.156714))'
+  ) AS geom
+),
+pts AS (
+  SELECT * FROM (VALUES
+    (-123.1061120::DOUBLE, 48.144285::DOUBLE)
+  ) AS t(lon, lat)
+)
+SELECT
+  ST_XMin(geom) AS xmin,
+  ST_XMax(geom) AS xmax,
+  ST_Contains(geom::POLYGON_2D, ST_MakePoint(lon, lat)::POINT_2D) AS typed_contains,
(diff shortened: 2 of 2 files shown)


Maxxen:
Thanks! This function should probably be removed/reworked eventually though as its not actually faster than the GEOMETRY equivalent once the geometries get big enough...

oglego:
Thanks again for the review, and for the heads up on that, that is good to know!

Issue #706, opened 2025-11-05
Is RTree index thread safe?


DvirDukhan:
Hi
I'm trying to use the duckdb spatial extension in a multi-threaded-single-process pattern, with duckdb-rs rust client.
AFAIU, duckdb allows inner process concurrency as long as there are no write conflicts.
My flow is appending new records (with new primary key) to a table that is indexed by RTree.

I have a single connection which I share between the threads
```rust
pub struct Database {
    db: Mutex<Connection>,
}

impl Database {
.....

pub fn insert(....)  {  // this function is executed concurrently
    let connection = {
        let g = self.db.lock()?;  // lock only for cloning connections in threads
        g.try_clone()?
    };
    let sql = "INSERT INTO .....";  // This is where we append to the table and the index should be updated.
    let exec_res = connection.execute(
        &sql,
        params![....]);
   // Handle exec_res Ok or Err
}
```
From time to time I'm getting segmentation faults with the following stack trace:
``` sh
#0 in void std::__introsort_loop<duckdb::RTreeEntry*, long, __gnu_cxx::__ops::_Iter_comp_iter<duckdb::RTreeNode::SortEntriesByRowId()::{lambda(duckdb::RTreeEntry const&, duckdb::RTreeEntry const&)#1}> >(duckdb::RTreeEntry*, duckdb::RTreeEntry*, long, __gnu_cxx::__ops::_Iter_comp_iter<duckdb::RTreeNode::SortEntriesByRowId()::{lambda(duckdb::RTreeEntry const&, duckdb::RTreeEntry const&)#1}>) () from /opt/duckdb/extensions/v1.4.0/linux_amd64/spatial.duckdb_extension
#1 in duckdb::RTree::SplitNode(duckdb::RTreeEntry&) const () from /opt/duckdb/extensions/v1.4.0/linux_amd64/spatial.duckdb_extension
#2 in duckdb::RTree::BranchInsert(duckdb::RTreeEntry&, duckdb::RTreeEntry const&) () from /opt/duckdb/extensions/v1.4.0/linux_amd64/spatial.duckdb_extension
#3 in duckdb::RTree::BranchInsert(duckdb::RTreeEntry&, duckdb::RTreeEntry const&) () from /opt/duckdb/extensions/v1.4.0/linux_amd64/spatial.duckdb_extension
[...]

PR #862, opened 2026-09-02
fix: handle non-flat row id vectors when replaying RTREE index appends and deletes from the WAL

### LINKED ISSUE #861: RTREE index: WAL-replayed appends fail with "Operation requires a flat vector", leaving the database permanently uncheckpointable
### What happens?

If a database file is closed without checkpointing while its WAL contains inserts into a table with an `RTREE` index, then on the next open:

- any statement that binds the index fails with `INTERNAL Error: Operation requires a flat vector but a non-flat vector was encountered`
- `CHECKPOINT` fails with `FATAL Error: Failed to create checkpoint because of error: Operation requires a flat vector but a non-flat vector was encountered`, after which DuckDB reports the database as invalidated

The state doesn't appear to be self-clearing, and so every subsequent open replays the WAL and hits the same failure.

I think a single inserted row is enough to reproduce.

### To Reproduce

Two `duckdb` CLI invocations against the same file.

```sql
-- Session 1
PRAGMA disable_checkpoint_on_shutdown;
CREATE TABLE t (geom GEOMETRY);
CREATE INDEX i ON t USING RTREE (geom);
INSERT INTO t VALUES (ST_Point(0,0));
```

```sql
-- Session 2, same file
LOAD spatial;
CHECKPOINT;
-- FATAL Error: Failed to create checkpoint because of error:
--   Operation requires a flat vector but a non-flat vector was encountered
```

Or, for the other symptom:

```sql
-- Session 2, same file
LOAD spatial;
INSERT INTO t VALUES (ST_Point(1,1));
-- INTERNAL Error: Operation requires a flat vector but a non-flat vector was encountered
```

`disable_checkpoint_on_shutdown` is only there to make the WAL survive deterministically. In the CLI a kill races an automatic checkpoint and usually loses, but I've seen the same state arise from a long-running process being killed mid-write. Either way the on-disk state is the same, and it's the state at open time that matters here: a database file plus a WAL of committed transactions, which is what any uncheckpointed shutdown leaves.

### Environment

- **OS:** macOS 26.6.2, arm64
- **DuckDB Version:** v1.5.5
- **DuckDB Client:** CLI
- **Hardware:** Apple Silicon
- **Spatial extension:** `eb1e57c`, `install_mode = REPOSITORY`

I don't think it's specific to the platform or the version. duckdb/duckdb-rs#733 reports the same error and the same stack on v1.5.1, linux_amd64.

### What isolates it

Same two-session structure, varying only the setup:

| Session 1 setup                                    | `CHECKPOINT` in session 2 |
| -------------------------------------------------- | ------------------------- |
| `RTREE` index, then 1 row inserted                 | **fails**                 |
| No index at all                                    | ok                        |
| ART index on an `INTEGER` column                   | ok                        |
| `RTREE` index, no rows inserted                    | ok                        |
| 1 row inserted, **then** the `RTREE` index created | ok                        |

So it's not index creation, and not the buffered-replay machinery in general. It seems like an ART index over the same shape of workload replays fine. It's specifically appends into an already-existing `RTREE` index, replayed from the WAL.

### Scope of the damage

The data itself seems fine, which might help narrow it down:

- `SELECT count(*) FROM t` returns every row, including the ones only in the WAL
- a read-only open (`duckdb -readonly`) reads everything
- only index-bound access and `CHECKPOINT` fail

### Workaround

Dropping the index lets the checkpoint through, and no rows are lost:

```sql
LOAD spatial;
DROP INDEX i;
CHECKPOINT;          -- succeeds, WAL is cleared
CREATE INDEX i ON t USING RTREE (geom);   -- index rebuilds fine afterwards
```

`CHECKPOINT` on its own and `FORCE CHECKPOINT` both fail. The index has to go first.

### Stack trace

```
INTERNAL Error: Operation requires a flat vector but a non-flat vector was encountered

0  duckdb::Exception::ToJSON(duckdb::ExceptionType, std::string const&)
1  duckdb::InternalException::InternalException(std::string const&)
2  duckdb::FlatVector::VerifyFlatVector(duckdb::Vector const&)
3  duckdb::RTreeIndex::Insert(duckdb::IndexLock&, duckdb::DataChunk&, duckdb::Vector&)
4  duckdb::RTreeIndex::Append(duckdb::IndexLock&, duckdb::DataChunk&, duckdb::Vector&)
5  duckdb::BoundIndex::Append(duckdb::DataChunk&, duckdb::Vector&, duckdb::IndexAppendInfo&)
6  duckdb::BoundIndex::ApplyBufferedReplays(duckdb::vector<duckdb::LogicalType, ...> const&,
                                            duckdb::BufferedIndexReplays&,
                                            duckdb::vector<duckdb::StorageIndex, ...> const&)
7  duckdb::TableIndexList::Bind(duckdb::ClientContext&, duckdb::DataTableInfo&, char const*)
```

### Impact

Any application holding a persistent DuckDB file with an `RTREE` index ends up with a database it can't checkpoint again after an unclean shutdown, and with index-bound queries failing, until someone drops the index by hand. The failure only shows up on the next open, so it lands during a restart rather than at the time of the write.

### Related

- duckdb/duckdb-rs#733 -- same error, same stack, reported against v1.5.1. It was filed on the Rust binding repo, and a maintainer said there in March 2026 that it looked like a spatial issue, but I couldn't find a corresponding issue here. The reproduction in that one goes through Rust and deliberately leaks a connection, which I think made it look like a connection-lifetime problem rather than an uncheckpointed WAL. Hence the CLI-only version above.
- duckdb/duckdb-spatial#706 -- segfaults in `RTreeIndex::Append` under concurrent inserts. Different symptom, same function, so possibly the same root cause.

### Reporter

- **Full Name:** Darren Spurgeon
- **Affiliation:** Independent

---

- [x] Did you include all relevant configuration to reproduce the issue?
- [x] Did you include all code required to reproduce the issue?
- [x] Did you include all relevant data sets for reproducing the issue? (the repro creates its own)

darrenspurgeon:
Fixes #861.

### Problem

If a database file is closed without checkpointing while its WAL contains inserts into a table with an `RTREE` index, then on the next open:

- any statement that binds the index fails with `INTERNAL Error: Operation requires a flat vector but a non-flat vector was encountered`
- `CHECKPOINT` fails with the same error, after which DuckDB reports the database as invalidated

It doesn't appear to clear itself. Every subsequent open replays the same WAL and hits the same failure. The data itself seems intact — `SELECT count(*)` returns every row and a read-only open reads everything — but the file can't checkpoint again until someone drops the index by hand.

A single inserted row is enough to reproduce:

```sql
-- Session 1
PRAGMA disable_checkpoint_on_shutdown;
LOAD spatial;
CREATE TABLE t (geom GEOMETRY);
CREATE INDEX i ON t USING RTREE (geom);
INSERT INTO t VALUES (ST_Point(0,0));
```

```sql
-- Session 2, same file
LOAD spatial;
CHECKPOINT;
-- FATAL Error: Failed to create checkpoint because of error:
--   Operation requires a flat vector but a non-flat vector was encountered
```

### Root cause

When the WAL is replayed without the spatial extension loaded, the `RTREE` index cannot be bound, so DuckDB buffers the appends and deletes and applies them later after something binds the index. `BoundIndex::ApplyBufferedReplays` slices the buffered chunk to build the row id vector:

```cpp
SelectionVector sel(offset_in_chunk, rows_to_process);
...
Vector row_ids(state.current_chunk.data.back(), sel, rows_to_process);
const auto error = Append(table_chunk, row_ids, append_info);
```

That produces a dictionary vector. `ConvertToEntries` read it with `FlatVector::Validity` and `FlatVector::GetData`, which throw on anything non-flat.

ART indexes may survive the same replay path because `ART::GenerateKeyVectors` goes through `UnifiedVectorFormat`, which is why the failure should be specific to `RTREE`.

### Fix

- `ConvertToEntries` reads the row ids through `UnifiedVectorFormat`, indexing both the data and the validity mask through the selection vector. This covers `Insert` and `Delete`, which share the helper.
- `key_chunk.SetCardinality(count)` before `key_chunk.Flatten()` in `Insert` and `Delete`. `DataChunk::Flatten()` flattens up to `size()`, which is 0 right after `Reset()`, so the flatten didn't do anything. I don't see a known trigger, since the key vector comes back flat on both paths, but the call was not doing what it reads like it should be doing.

### Test

`test/sql/index/rtree_persistence_wal_append.test` covers both the append and the delete replay: create the index, write, restart, then index scan and `CHECKPOINT`. It fails before this change and passes after.

Two notes on the test:

- The existing `rtree_persistence_wal.test` didn't look like it was actually exercising WAL replay. The sqllogictest runner defaults `checkpoint_wal_size` to 0 (`TestConfiguration::GetCheckpointWALSize`), so it checkpoints as soon as the WAL is non-empty and no `.wal` file survives the `restart`. The new test raises `checkpoint_threshold` after every restart, which is what keeps the WAL alive into the next session. I left the existing test alone, but certainly they can be merged.
- The delete assertions read the index directly with `rtree_index_dump` instead of only counting table rows. Deleted rows are filtered out when the index scan fetches them from the table, so a table-level count passes whether or not the entries were actually removed from the index.

### Related

- duckdb/duckdb-rs#733 looks to be the same error and same stack, reported against v1.5.1 on linux_amd64. Filed on the Rust binding repo; the reproduction there goes through Rust and deliberately leaks a connection, which made it look like a connection-lifetime problem rather than an uncheckpointed WAL.
- #706 appears to be unrelated despite also involving `RTreeIndex::Append`. Its stack crashes inside `RTree::SplitNode` / `SortEntriesByRowId` on the ordinary `LocalTableStorage::AppendToIndexes` path, which points at concurrent tree mutation rather than vector layout.

### DIFF
--- a/src/spatial/index/rtree/rtree_index.cpp
+++ b/src/spatial/index/rtree/rtree_index.cpp
@@ -225,7 +225,6 @@ void RTreeIndex::CommitDrop(IndexLock &index_lock) {
 template <class CALLBACK = std::function<void(const RTreeEntry &)>>
 static void ConvertToEntries(Vector &box_vec, Vector &rowid_vec, idx_t count, CALLBACK &&callback) {
 	const auto &box_validity = FlatVector::Validity(box_vec);
-	const auto &row_validity = FlatVector::Validity(rowid_vec);
 
 	const auto &box_entries = StructVector::GetEntries(box_vec);
 	const auto box_xmin_data = FlatVector::GetData<float>(*box_entries[0]);
@@ -233,10 +232,13 @@ static void ConvertToEntries(Vector &box_vec, Vector &rowid_vec, idx_t count, CA
 	const auto box_xmax_data = FlatVector::GetData<float>(*box_entries[2]);
 	const auto box_ymax_data = FlatVector::GetData<float>(*box_entries[3]);
 
-	const auto row_data = FlatVector::GetData<row_t>(rowid_vec);
+	UnifiedVectorFormat rowid_format;
+	rowid_vec.ToUnifiedFormat(count, rowid_format);
+	const auto row_data = UnifiedVectorFormat::GetData<row_t>(rowid_format);
 
 	for (idx_t i = 0; i < count; i++) {
-		if (!box_validity.RowIsValid(i) || !row_validity.RowIsValid(i)) {
+		const auto row_idx = rowid_format.sel->get_index(i);
+		if (!box_validity.RowIsValid(i) || !rowid_format.validity.RowIsValid(row_idx)) {
 			continue;
 		}
 
@@ -246,7 +248,7 @@ static void ConvertToEntries(Vector &box_vec, Vector &rowid_vec, idx_t count, CA
 		box.max.x = box_xmax_data[i];
 		box.max.y = box_ymax_data[i];
 
-		const auto row = row_data[i];
+		const auto row = row_data[row_idx];
 
 		RTreeEntry new_entry = {RTree::MakeRowId(row), box};
 
@@ -256,14 +258,16 @@ static void ConvertToEntries(Vector &box_vec, Vector &rowid_vec, idx_t count, CA
 }
 
 ErrorData RTreeIndex::Insert(IndexLock &lock, DataChunk &input, Vector &row_vec) {
+	const auto count = input.size();
 
 	key_chunk.Reset();
 	key_executor->ExecuteExpression(input, key_chunk.data[0]);
+	key_chunk.SetCardinality(count);
 	key_chunk.Flatten();
 
 	auto &box_vec = key_chunk.data[0];
 
-	ConvertToEntries(box_vec, row_vec, input.size(), [&](const RTreeEntry &entry) { tree->Insert(entry); });
+	ConvertToEntries(box_vec, row_vec, count, [&](const RTreeEntry &entry) { tree->Insert(entry); });
 
 	return ErrorData {};
 }
@@ -284,6 +288,7 @@ void RTreeIndex::Delete(IndexLock &lock, DataChunk &input, Vector &row_vec) {
 
 	key_chunk.Reset();
 	key_executor->ExecuteExpression(expr_chunk, key_chunk.data[0]);
+	key_chunk.SetCardinality(count);
 	key_chunk.Flatten();
 
 	auto &box_vec = key_chunk.data[0];
--- a/test/sql/index/rtree_persistence_wal_append.test
+++ b/test/sql/index/rtree_persistence_wal_append.test
@@ -1,0 +1,106 @@
+# name: test/sql/index/rtree_persistence_wal_append.test
+# group: [index]
+
+# Appends and deletes on an existing RTREE index are buffered during WAL replay (when spatial isn't
+# loaded) and applied when the index is bound. Row ids arrive as a dictionary vector, which used to
+# error with "Operation requires a flat vector", making the database uncheckpointable.
+# See https://github.com/duckdb/duckdb-spatial/issues/861
+#
+# The runner checkpoints whenever the WAL is non-empty, so 'checkpoint_threshold' must be raised
+# after each restart to keep the WAL alive.
+
+require spatial
+
+load __TEST_DIR__/rtree_persistence_wal_append.db
+
+statement ok
+SET checkpoint_threshold='1GB';
+
+statement ok
+PRAGMA disable_checkpoint_on_shutdown;
+
+statement ok
+CREATE TABLE t1 (geom GEOMETRY);
+
+statement ok
+CREATE INDEX my_idx ON t1 USING RTREE (geom);
+
+# Insert after index creation so the appends go through the WAL
+statement ok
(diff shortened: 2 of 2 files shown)


src/spatial/index/rtree/rtree_index.cpp
```diff
@@ -256,14 +258,16 @@ static void ConvertToEntries(Vector &box_vec, Vector &rowid_vec, idx_t count, CA
 }
 
 ErrorData RTreeIndex::Insert(IndexLock &lock, DataChunk &input, Vector &row_vec) {
+	const auto count = input.size();
 
 	key_chunk.Reset();
 	key_executor->ExecuteExpression(input, key_chunk.data[0]);
+	key_chunk.SetCardinality(count);
 	key_chunk.Flatten();
```
Maxxen, line 266: This doesn't really make sense to me - calling `.Flatten()` here should ensure the `box_vec` is a flat vector, making ConvertToEntries safe as is

Maxxen:
Hello! Thanks for digging into this and thanks for this PR!
Nevermind the above comment - this seems correct to me.

PR #867, opened 2026-09-08
fix(serde): raise catchable InvalidInputException instead of internal Buffer overflow on truncated/empty WKB (#866)

### LINKED ISSUE #866: INTERNAL Error: Buffer overflow in Serde::Deserialize on a zero-length WKB blob (ST_GeometryType, ST_IsEmpty, ST_X and others; ST_AsText handles it correctly)
**DuckDB:** v1.5.5 (macOS arm64, Homebrew). **spatial:** `eb1e57c` (core repository).

A zero-length `BLOB` passed through `ST_GeomFromWKB` aborts the query with an internal assertion instead of raising a catchable error, when the geometry is consumed in column context:

```sql
CREATE TABLE g AS SELECT ''::BLOB AS w;
SELECT ST_GeometryType(ST_GeomFromWKB(w)) FROM g;
```

```
INTERNAL Error: Buffer overflow
  duckdb::Serde::Deserialize(sgl::geometry&, duckdb::ArenaAllocator&, char const*, unsigned long) + 888
  duckdb::(anonymous namespace)::ST_GeometryType::ExecuteGeometry(...) + 1280
  duckdb::ExpressionExecutor::Execute(duckdb::BoundFunctionExpression const&, ...)
  duckdb::PhysicalProjection::Execute(...)
```

`TRY()` does not help, which is expected — an internal error is not catchable — but it does mean there is no way to tolerate this input at the SQL level.

### The project's own behaviour says what the contract should be

`ST_AsText` on the identical input raises an ordinary, catchable error:

```
Invalid Input Error: Unexpected end of binary data at position 0
```

So this is not "garbage in, garbage out". One code path validates and reports; the others assert.

### Which functions are affected

Same table, same `ST_GeomFromWKB(w)`, different consumer, on 1.5.5:

| consumer | result |
| --- | --- |
| `ST_AsText` | `Invalid Input Error: Unexpected end of binary data at position 0` (catchable) |
| `ST_AsWKB` | no error — returns a 0-byte blob |
| `ST_GeometryType` | **INTERNAL Error: Buffer overflow** |
| `ST_IsEmpty` | **INTERNAL Error: Buffer overflow** |
| `ST_Dimension` | **INTERNAL Error: Buffer overflow** |
| `ST_NPoints` | **INTERNAL Error: Buffer overflow** |
| `ST_X` | **INTERNAL Error: Buffer overflow** |
| `ST_Force2D` | **INTERNAL Error: Buffer overflow** |

`ST_AsWKB` round-tripping the empty blob back out at 0 bytes suggests `ST_GeomFromWKB` stores the input without validating it, and each consumer decides for itself whether to check the length before deserializing.

### It only reproduces in column context

As a folded constant the same expression raises a normal catchable error, so a scalar reproduction will not show it:

```sql
SELECT ST_GeometryType(ST_GeomFromWKB(''::BLOB));
-- Invalid Input Error: Unexpected end of binary data at position 0
```

A *truncated* (rather than empty) blob is also handled correctly — `unhex('010100')` gives `Invalid Input Error: Unexpected end of binary data at position 1`. Zero length is the case that asserts.

### Why it matters

The trigger is data-dependent, so it fires in production on malformed third-party input rather than in testing — which is exactly the input a validation pipeline exists to find. Because the error is internal rather than catchable, there is no query-level way to skip the bad rows.

**Workaround** for anyone who lands here — gate on length first, which short-circuits at vector level:

```sql
CASE WHEN octet_length(wkb) >= 5 THEN ST_GeometryType(ST_GeomFromWKB(wkb)) END
```

### Related

#792 was the same class (`INTERNAL Error: Buffer overflow` out of a spatial function, `TRY()` unable to catch it) via `ST_AsMVTGeom` + `ST_IsEmpty`, and was fixed in 1.5.2. This entry point is still live on 1.5.5. #858 was a different mechanism (a core `OPERATOR_TRY` selection-vector bug, fixed in 1.5.2) and is not involved here — 1.5.5 already carries that fix.

connerkup:
### Summary

Fixes #866

In column context, passing a 0-length blob (`''::BLOB`) through `ST_GeomFromWKB(w)` returns an empty 0-byte geometry blob without failing early. When subsequent vector functions (e.g. `ST_GeometryType`, `ST_IsEmpty`, `ST_Dimension`, `ST_NPoints`, `ST_X`, `ST_Force2D`) deserialize the geometry, `BinaryReader::CheckSize` throws an uncatchable `INTERNAL Error: Buffer overflow` assertion:

```sql
CREATE TABLE g AS SELECT ''::BLOB AS w;
SELECT ST_GeometryType(ST_GeomFromWKB(w)) FROM g;
-- INTERNAL Error: Buffer overflow
```

Because an `InternalException` represents an assertion failure within DuckDB rather than user input error, it cannot be caught via SQL `TRY()` and immediately aborts the query and invalidates the client connection. Furthermore, `ST_AsText` on the identical input raises a standard, catchable error:
```sql
SELECT ST_AsText(ST_GeomFromWKB(w)) FROM g;
-- Invalid Input Error: Unexpected end of binary data at position 0
```

### Root Cause

1. In `src/spatial/util/binary_reader.hpp`, `BinaryReader::CheckSize` threw `InternalException("Buffer overflow")` when `ptr + size > end`. A buffer underrun while reading deserialized geometry data is an invalid/truncated input condition, not an internal database assertion failure.
2. In `src/spatial/geometry/geometry_serialization.cpp`, `DeserializeInternal` had no guard for empty buffers (`buffer_size == 0`), immediately triggering `CheckSize(sizeof(uint8_t))` at offset 0.

### Changes

1. **`src/spatial/util/binary_reader.hpp`**:
   - Updated `BinaryReader::CheckSize` to compare `size > static_cast<size_t>(end - ptr)` (preventing integer overflow on malicious sizes) and throw:
     ```cpp
     throw InvalidInputException("Unexpected end of binary data at position %zu",
                                 static_cast<size_t>(ptr - beg));
     ```
     This matches the exact error message and offset formatting used by DuckDB's core `BlobReader`.
2. **`src/spatial/geometry/geometry_serialization.cpp`**:
   - Added pre-flight check in `DeserializeInternal` for `buffer_size == 0` to immediately raise `InvalidInputException("Unexpected end of binary data at position 0")`.
3. **`test/sql/geometry/st_geomfromwkb.test`**:
   - Added regression tests verifying that zero-length WKB blobs in column context raise catchable `Invalid Input Error: Unexpected end of binary data at position 0` across `ST_GeometryType`, `ST_IsEmpty`, `ST_Dimension`, `ST_NPoints`, and `ST_X`.
   - Verified that `TRY(ST_GeometryType(...))` and `TRY(ST_IsEmpty(...))` catch the error and cleanly return `NULL`.


Maxxen:
Thanks!

We should probably just fix `ST_GeomFromWKB` in core to reject zero sized inputs anyway, so that there is no risk of creating zero sized GEOMETRY that can leak into the rest of the system.

